PR #22, opened 2016-11-05
Fix failing systemd ssh.service

jkirk:
SSH needs host keys which previously were generated by
ssh-keygen.service. But the generation of the host keys is not a
"service" and must only be done when no host keys exist.

Also ssh-bootoption.service was aliased with sshd.service which itself
depends on the kernel option "ssh" which prevents starting of ssh. So I
removed the alias entry.


jkirk:
This is just a possible starting point. Obviously we also need a working ssh-bootoption.service but that depends on working network setup on boot time.

etc/grml/fai/config/files/etc/systemd/system/ssh.service/GRMLBASE
```diff
@@ -0,0 +1,24 @@
+# This file was deployed via grml-live's
+# ${GRML_FAI_CONFIG}/config/scripts/GRMLBASE/15-initsetup, using
+# ${GRML_FAI_CONFIG}/config/files/etc/systemd/system/ssh-bootoption.service/GRMLBASE
```
mika: This filename doesn't match the actual filename here. :)

etc/grml/fai/config/files/etc/systemd/system/ssh.service/GRMLBASE
```diff
@@ -0,0 +1,24 @@
...
+Description=OpenBSD Secure Shell server
+After=network.target auditd.service
+ConditionPathExists=!/etc/ssh/sshd_not_to_be_run
+
+[Service]
+EnvironmentFile=-/etc/default/ssh
+ExecStartPre=-/bin/sh -c "[ ! -f /etc/ssh/ssh_host_rsa_key ] && /usr/bin/ssh-keygen -t rsa     -f /etc/ssh/ssh_host_rsa_key     -C '' -N ''"
+ExecStartPre=-/bin/sh -c "[ ! -f /etc/ssh/ssh_host_dsa_key ] && /usr/bin/ssh-keygen -t dsa     -f /etc/ssh/ssh_host_dsa_key     -C '' -N ''"
+ExecStartPre=-/bin/sh -c "[ ! -f /etc/ssh/ssh_host_ecdsa_key ] && /usr/bin/ssh-keygen -t ecdsa   -f /etc/ssh/ssh_host_ecdsa_key   -C '' -N ''"
+ExecStartPre=-/bin/sh -c "[ ! -f /etc/ssh/ssh_host_ed25519_key ] && /usr/bin/ssh-keygen -t ed25519 -f /etc/ssh/ssh_host_ed25519_key -C '' -N ''"
```
mika: Maybe it makes sense to indent the lines as it used to be done in etc/grml/fai/config/files/etc/systemd/system/ssh-keygen.service/GRMLBASE  before (see above)?

mika: Jey, thanks for the PR! Just to clarify, we want to ship our own ssh.service and not name it ssh-bootoption.service or such, right?

etc/grml/fai/config/files/etc/systemd/system/ssh.service/GRMLBASE
jkirk: Ha! Yes, you are completely right! I will fix that!

mika:
Thanks! :+1:

PR #23, opened 2016-11-19
Enable and disable services by using systemd.presets

jkirk:
To enabled or disable services on boot-time we should use systemd.presets. This change implements the effort of `15-initsetup` in `/etc/systemd/system/system-preset/10-grml.preset`.

And: There is also a "minor" fix where `fcopy` did not set ownership/permission correctly when copying `/etc/systemd`.


mika:
Thanks! :+1:

PR #25, opened 2017-01-15
Add SHA-256 and SHA-512 hashes

bk2204:
Most security-conscious users expect to find SHA-256 hashes of files, since both MD5 and SHA-1 are considered weak.  SHA-512 additionally has the benefit that it is significantly faster than SHA-256 on 64-bit systems, processing twice as much data per chunk with only 125% of the operations.  Generate both SHA-256 and SHA-512 hashes anywhere we provide SHA-1 hashes.


mika:
Sorry for the delay, merged - thanks for the PR! :+1:

PR #27, opened 2017-01-27
Change systemd default target to multi-user.target

jkirk:
The default target used to be graphical.target which is not appropriate
for Grml


mika:
Merged, thanks! :+1:

PR #28, opened 2017-01-27
Use ssh-keygen -A to generate host keys

jkirk:
ssh-keygen -A (from manpage):
  For each of the key types (rsa1, rsa, dsa, ecdsa and
  ed25519) for which host keys do not exist, generate the host keys
  with the default key file path, an empty passphrase, default bits for
  the key type, and default comment.  This is used by system
  administration scripts to generate new host keys.


mika:
Merged, thanks! :+1:

Issue #31, opened 2017-03-15
do not write to / by default


anarcat:
i am not sure why the default output is `/grml`, but it seems a bit problematic to me. `/` is often quite tight in terms of space - it's not the place where there's the most storage, out of the box. in some cases, it's also readonly and may ring all sorts of alarm bells when ran.

better defaults could be:

* `./grml` - assume the current directory is a sane location
* `/srv/grml` - similar to `/grml` but at least respects the FHS somehow, and `/srv` often has lots of storage
* `/var/lib/grml` - probably the most conformant location...

The default seems to be set in https://github.com/grml/grml-live/blob/fda69fdbf5232e28a8536317656e16fca2ad2962/grml-live#L405 - i would make a PR, but since the change is trivial, i figured i would discuss it first. ;)

My prefered option is `./grml`, it's what i expected, not knowing much about the software.

anarcat:
aand just now i ran out of space building grml in `/grml`, while my `$HOME` is empty... ;)

mika:
Thanks for taking care of reporting such issues, I'm blind for such stuff since I'm using it for way too long. :)

I agree, /grml is a bad decision. I'd prefer either "./grml/" (current working directory) or "/srv/grml/", so let's go with current working directory "./grml/"! If you make a PR please don't forget updating docs (docs/grml-live.txt), thanks!

anarcat:
thanks for your openness! :)

PR #32, opened 2017-03-15
output to current directory instead of root

anarcat:
this would create problems as / is often space restricted

Closes: #31


mika:
Thanks, I've merged your PR with a slightly adjusted commit message. :+1:

PR #34, opened 2017-03-19
add stressant package into Grml

anarcat:
to quote from the control file:

> Stressant is a simple stress testing and burn-in tool
> 
> It is designed to run on new machines to make sure they will work
> reliably by testing various parts of the system (CPU, RAM, disk,
> network) by putting them under heavy load and try to detect failures.
> 
> As much as possible, stressant tries to reuse existing tools to
> perform the various tasks and aims to be run automatically.

it has just entered Debian sid and will pull at least 3 new
dependencies in (python-humanize, python-colorlog and stress-ng)

adding this is essential for the Stressant project to continue
collaborating with Grml.

### DIFF
--- a/etc/grml/fai/config/package_config/GRML_FULL
+++ b/etc/grml/fai/config/package_config/GRML_FULL
@@ -286,6 +286,7 @@ flashrom
 mcollective
 memtester
 puppet
+stressant
 
 # virtualization support
 imvirt


anarcat:
ping? i have added a manpage to stressant to comply with the [policy](http://grml.org/policy/) and I believe the [documentation](http://stressant.readthedocs.io/en/latest/) is pretty complete, if maybe a little minimal...

mika:
I didn't forget about it, sorry for not responding until now: I'd like to finish the new Grml stable release and avoid any non-positive test results if possible, though stressant isn't available in Debian/testing so it will raise a test warning on our builds.

anarcat:
Okay no problem! Thanks for the response... It would be great to see
stressant shipped with the newest grml release! :)

mika:
Grml is based on Debian/testing, JFTR :)

anarcat:
On 2017-04-27 07:34:01, Michael Prokop wrote:
> Grml is based on Debian/testing, JFTR :)

but there are builds for unstable as well, no?

mika:
Yes, but the package selection (currently) is independent of the Debian release, so whatever is in the list is trying to be installed.

anarcat:
i see... will there be a possibility of including sid-only packages later? or through backports once buster hits the rails? :)

mika:
Now having the new Grml stable release out and stretch soon to be released (so stressant will migrate to Debian/testing) I've looked in this and noticed the following for grml-full 2017.05 (adding unstable to `sources.list`):

```
The following NEW packages will be installed:
  fonts-font-awesome fonts-lato libjs-jquery libjs-modernizr libjs-sphinxdoc libjs-underscore python-colorlog python-humanize sphinx-rtd-theme-common stressant
0 upgraded, 10 newly installed, 0 to remove and 141 not upgraded.
Need to get 3,869 kB of archives.
After this operation, 15.7 MB of additional disk space will be used.
```

What's the rationale for the fonts, libjs and sphinx* packages here? TBH, 15.7MB is quite heavy for 284.0 kB of installed size for stressant itself. :(

anarcat:
Ouch!

That's probably the documentation... I will need to split that out in a
stressant-doc package, I guess...

anarcat:
should be fixed in the pending 0.4.1 upload (which may have to pass NEW because of the -doc package addition).

anarcat:
stressant 0.4.1 has now passed NEW.

mika:
I've just merged this, thanks for the PR :)

PR #37, opened 2017-05-19
Issues/#35

thomasdstewart:
To fix https://github.com/grml/grml/issues/35


mika:
Merged squashed and slightly reworked, thanks for the patch! \o/ :+1:

PR #41, opened 2017-05-21
Get rid of grml-runtty

jkirk:
What once was done by grml-runtty can be now be accomplished by systemd.

That said, there is a weird behavior when starting GNU/screen via
"run-screen". When started via grml-runtty screen runs zsh, but
when started directly via systemd screen runs /bin/sh (= dash).

The reason for that is, that the SHELL variable is empty when called
directly via systemd whereas SHELL=/bin/zsh when started via
grml-runtty. I could not figure out why but a solution would be to set
the "Environment"-option in the systemd-unit which is proposed in a
separate commit.

(partly) fixes grml/grml#14


mika:
Great work, thanks! :+1:

PR #42, opened 2017-05-31
Update bootprompt / cheatcodes documentation

jkirk:
The stats feature was removed, therefore we also do not need the nostats
option anymore.

The stats feature was removed in grml-autconfig-commit# 7138a24fb

Partly fixes grml/grml#9


mika:
All merged, great work - thanks :) :+1:

PR #43, opened 2017-06-07
Regroup package list of GRML_SMALL

jkirk:
To make it easier to spot differences of GRML_FULL and GRML_SMALL I
resorted the package list like in GRML_FULL. Another alternative would
be to sort both lists alphabetically. Either way we should be
consistent.

### DIFF
--- a/etc/grml/fai/config/package_config/GRML_FULL
+++ b/etc/grml/fai/config/package_config/GRML_FULL
@@ -1,5 +1,6 @@
 PACKAGES install
 
+# grml
 grml-live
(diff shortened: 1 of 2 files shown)


jkirk:
Please note that I added comments like "# missing in full" and "# added on merge" on some lines:
```
ed # missing in full
tcpd # missing in full
grml-etc-core # was added on merge
```
Please remove the comment and/or the line of subject if its by intention or by accident.

jkirk:
Canceled regrouping.

Issue #46, opened 2017-11-08
Cannot build grml from inside grml


Zugschlus:
This is a reminder ticket for an issue that has been discussed on the mailing list to avoid that the issue gets forgotten.

I am trying rebuild a grml iso from within grml with
sudo grml-live -A -V -s sid -c DEBORPHAN,GRMLBASE,GRML_FULL,RELEASE,AMD64,IGNORE,SNAPSHOT -r "test20170915" -g
grml64 -o ~/grml-remaster/tmp
and having this fail with a message that /efi/boot/bootx64.efi is not found at the place where the script its looking for it, while the file is in /ilb/live/mount/medium.
mika has shown a pach on the mailing list, but that patch doesn'thelp.
Full thread reference is starting at http://ml.grml.org/pipermail/grml/2017-September/011696.html

Greetings
Marc

R-Sommer:
Just tried it successfully with Grml 2022.11 (with a workaround of failing /etc/grml/fai/config/scripts/GRMLBASE/91-update-pciids - which I wasn't able to figure out).

mika:
@R-Sommer ok great, thanks for the feedback! Regarding the failure in `/etc/grml/fai/config/scripts/GRMLBASE/91-update-pciids`: this usually happens if the upstream website isn't available. Executing this script can be disabled via the Grml class `NO_ONLINE` during grml-live execution.

I consider this as resolved, if anyone should have any further problems please comment or open a new issue. Thx!

R-Sommer:
Thanks @mika, the failure was due to missing proxy configuration. After adding proxy settings, I could make a successful try.

Issue #45, opened 2017-11-08
please consider a possibility to switch off the graphical startup screen by patching the ISO


Zugschlus:
I regularly use a grml USB stick in a server's USB port to have a rescue system that is independent of both local disk and network. Of course, the easiest way to have this is just dd'ing an grml iso to
the medium and then plugging it in without more ado. In the past, I had to build my own grml iso to have a non-graphical startup to be able to select the grml startup option, since that one is usually shown with a fancy graphical isolinux screen and doesn't work if the only thing one has is a serial port.

I would love to have the possibility to do a simple patch in an otherwise unchanged grml image so that my ProLiant grml is as close to "release" grml as possible. I would appreciate a hint if that is
already possible.

While I do understand that having a nice-looking selection screen is nice and important for user acceptance, this severely reduces grml's attractivity for the server jockey who usually doesn't have graphical access to the systems.

A possible implementation would be to include a longish placeholder comment into the isolinux configuration that can be used to patch in additional commands in an existing iso. I am not afraid of a hex editor, so I'm fine with doing that.

Greetings
Marc

mika:
There already is https://github.com/grml/grml-live/blob/master/templates/boot/isolinux/default.cfg - does this help you?

Zugschlus:
The pointer towards the git templates was really helpful. I found a comment in https://github.com/grml/grml-live/blob/master/templates/boot/isolinux/isolinux.cfg:

```
# finally either include 'console.cfg' for console based bootsplash
# or 'vesamenu.cfg' for graphic menu version of bootsplash (adjusted
# automatically in build process via grml-live):
```

with that, I was able to use dhex on the iso image, to remove the # in front of the serial stanza and to replace include vesamenu.cfg by include console.cfg.

This is exactly what I need.

I would however suggest putting in placeholders as in default.cfg also in isolinux.cfg.

Zugschlus:
I have put instructions in the wiki, http://wiki.grml.org/doku.php?id=rescue
The issue is now reduced to "put in palceholers into isolinux.cfg" and can be closed then. Thanks for helping.

Zugschlus:
closed by merging PR #51

Issue #44, opened 2017-08-31
please consider giving information how an iso was built


Zugschlus:
Hi,
to debug something, I would like to be able to build my own grml-full iso that is as close as possible to the official image. Please consider putting information about how an image was built into the iso. Minimum is the combination of fai classes that were used, I would love to have the exact grml-live command line.
Can you please give the combination of fai classes used for grml-{full|small}-2017.05 either in this ticket or in e-mail, so that I can continue with my debugging?
Greetings
Marc

mika:
As written on grml mailing list the following classes are involved:

    DEBORPHAN,GRMLBASE,GRML_FULL,RELEASE,AMD64,IGNORE,SNAPSHOT

Zugschlus:
As we discussed weeks ago in IRC, would it be possible to have that information inside the grml iso? If this is actually already implemented, this issue can be closed now. It would be great if the path in the running grml that tells the user how the running grml was built was mentioned here for reference and the search engines.

mika:
So I implemented this feature via https://github.com/grml/grml-live/commit/9453222c016a82a2c964e174ec01687f88d73d86 for grml-live. Our daily Grml ISOs starting as of tomorrow (2021-07-17) are supposed to include this file, see `/run/live/medium/conf/buildinfo.json` on the running live system then.

Please let me know if that provides everything you're interested in.

FTR: we need to decide yet whether this will be used as-is for our next stable release.

PR #51, opened 2018-03-19
add placeholder and explanation inviting local patches to isolinux.cfg

Zugschlus:
this closes grml/grml-live #45

This is unfortunately completely untested, since to multiple issues, one
of them being grml/grml-live #44 and grml/grml-live #46 are still
unaddressed, I cannot currently reliably build grml images. As it is a
comment-only patch, this can be safely applied and I'll happily test a
daily image afterwards.

### DIFF
--- a/templates/boot/isolinux/isolinux.cfg
+++ b/templates/boot/isolinux/isolinux.cfg
@@ -76,6 +76,26 @@ timeout 300
 # use this to control the bootup via a serial port:
 # serial 0 9600
 
+# following is a placeholder just in case you want to patch your own startup
+# options right into the grml ISO. Feel free to use a hex editor like dhex to
+# adapt this part of the file inside the ISO to your needs. You can search for
+# the string a8bd612867ae71d0b17f9d33f80f61ab261aa354 in the ISO file binary to
+# find the right place
+
+# the following are 10 lines of 70 characters
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+
 # finally either include 'console.cfg' for console based bootsplash
 # or 'vesamenu.cfg' for graphic menu version of bootsplash (adjusted
 # automatically in build process via grml-live):
+
+


templates/boot/isolinux/isolinux.cfg
```diff
@@ -76,6 +76,26 @@ timeout 300
 # use this to control the bootup via a serial port:
 # serial 0 9600
 
+# following is a placeholder just in case you want to patch your own startup
+# options right into the grml ISO. Feel free to use a hex editor like dhex to
+# adapt this part of the file inside the ISO to your needs. You can search for
+# the string a8bd612867ae71d0b17f9d33f80f61ab261aa354 in the ISO file binary to
```
mika: What's the rational for `a8bd612867ae71d0b17f9d33f80f61ab261aa354`, can you please explain this a bit? Is this guaranteed to stay stable no-matter-what?

templates/boot/isolinux/isolinux.cfg
```diff
@@ -76,6 +76,26 @@ timeout 300
...
+######################################################################
+######################################################################
+######################################################################
+######################################################################
+
 # finally either include 'console.cfg' for console based bootsplash
 # or 'vesamenu.cfg' for graphic menu version of bootsplash (adjusted
 # automatically in build process via grml-live):
+
+
```
mika: Why those two empty lines here?

templates/boot/isolinux/isolinux.cfg
Zugschlus: This is just an arbitrary string that is unlike to show up at any other place inside the iso. I think it was the sha1 checksum of the current date. I don't remember what I fed into sha1sum to obtain that string. Replace with anything you like as long as it doesn't show up anywhere else.

templates/boot/isolinux/isolinux.cfg
Zugschlus: I think that's a left over from the editing process. I think I would remember a reason otherwise. Maybe just prettier reformatting for the file that actually ends up in the ISO?

templates/boot/isolinux/isolinux.cfg
mika: Ahhhhh, alright :) Thanks for clarifying, makes sens!

templates/boot/isolinux/isolinux.cfg
mika: It comes out like that:

```
# finally either include 'console.cfg' for console based bootsplash
# or 'vesamenu.cfg' for graphic menu version of bootsplash (adjusted
# automatically in build process via grml-live):
include vesamenu.cfg
```

And i'd prefer to keep it that way.

I'll slightly adjust your patch and will merge it then.

mika:
Merged the slightly adjusted PR, thanks! :+1:

Zugschlus:
I can confirm that this works as intended in the daily snapshot on 2018-08-06. I will write some docs in http://wiki.grml.org/doku.php?id=remastering
Thanks for merging.

PR #52, opened 2018-03-28
Use stretch-backports instead of jessie-backports for stable

HeadlessChild:
I assume that Debian stable should track Stretch in grml-live so then we need to track stretch-backports instead of jessie-backports for the DEBIAN_STABLE apt source file.


mika:
Nice catch, thanks for the PR! :+1:

PR #60, opened 2018-08-29
Update `grubx64.efi.signed`

paulmenzel:


paulmenzel:
This is just for testing.

PR #80, opened 2019-11-22
SW: cryptsetup-initramfs (fix persistence-encryption)

xdch47:
*cryptsetup in initramfs* has been moved to a [separate package](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=cryptsetup-initramfs), instead of setting `CRYPTSETUP=y` for the default initramfs scripts, since current stable debian *buster*. This is necessary to boot *grml* with `persistence-encryption=luks` as described [here](https://wiki.grml.org/doku.php?id=persistency).

### DIFF
--- a/etc/grml/fai/config/package_config/GRMLBASE
+++ b/etc/grml/fai/config/package_config/GRMLBASE
@@ -4,6 +4,7 @@ busybox
 bzip2
 console-data
 console-setup
+cryptsetup-initramfs
 dbus
 deborphan
 dmidecode


etc/grml/fai/config/package_config/GRMLBASE
```diff
@@ -33,6 +33,7 @@ haveged
 hdparm
 hwinfo
 initramfs-tools
+cryptsetup-initramfs
```
mika: Could you please add this to the list in a alphabetically sorted order?

etc/grml/fai/config/scripts/GRMLBASE/80-initramfs
```diff
@@ -20,8 +20,8 @@ fi
 echo "Rebuilding initramfs"
 
 for initrd in "$(basename $target/boot/vmlinuz-*)" ; do
-    if ! CRYPTSETUP=y $ROOTCMD update-initramfs -k "${initrd##vmlinuz-}" -c ; then
```
mika: I suppose we could just leave the `CRYPTSETUP=y` here to support building for older releases than buster as well, at least it shouldn't do any harm, nor?

xdch47: okay, will fix it, but I guess older release might fail, since the package `cryptsetup-initramfs` might not exists, or am I wrong?

xdch47:
added initramfs-crypt (sorted)

etc/grml/fai/config/scripts/GRMLBASE/80-initramfs
mika: I'm not entirely sure, but IIRC we had support for the crypto stuff already in older Debian releases supported by cryptsetup's hooks/scripts, before it split into a separate cryptsetup-initramfs package. Just want to make sure we don't surprise anyone building older Debian releases. :)

Thanks for adjusting this!

mika:
Thanks, merged and pushed. Thanks a ton! :+1:

xdch47:
Thank you guys :)

PR #81, opened 2019-11-22
Additional software

xdch47:
add `tree` and `ncdu` to `GRML_FULL`. IMHO they are quite useful and donot consume much space (might be candidates for `GRML_SMALL` as well?!).


mika:
Excellent suggestion, thanks for the PR, just merged!
I also added tree + ncdu to GRML_SMALL, since all the dependencies are there and the tools are small enough and useful to have available :)

Thanks!

PR #87, opened 2020-06-03
Add boot option legacynet (Legacy Network Interface Names)

jkirk:
Grml should use Predictable Network Interface Names by default.

To make that switch easier we added two new boot options:
"legacynet" and "pnet".

The boot option "legacynet" provides an easy way to use the old (legacy)
network interface names by adding "net.ifnames=0" to the kernel
command line, see: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

The kernel command line options "net.ifnames=0" currently in every other
boot option (except "pnet") but will be removed as soon as we switch to
Predictable Network Interface Names by default.

The boot options "pnet" is temporary (and therefor not documented
grml-cheatcodes.txt and in the isolinux help page) as this should just
help us boot Grml with Predictable Network Interface Names enabled.
As this will be our new default anyway, this boot options will be
removed as soon as the transition is completed.

See: grml/grml#127


jkirk:
As just discussed. We don't want a boot option "legacynet". Closing.

PR #94, opened 2020-12-27
GRUB: remove erroneous quotes around kernelopts

michaellass:
kernelopts are used by grml-rescueboot to communicate kernel arguments from the host's grub to the grub within the grml ISO. E.g., the host grub boot entry will contains something like:

````
kernelopts=" arg1 arg2 blacklist=foo etc "
export kernelopts
````

Currently, grub within the grml ISO will pass this entire string as one single argument to the kernel. Remove the quotes such that the arguments are interpreted as intended.

Fixes https://github.com/grml/grml-rescueboot/issues/5

### DIFF
--- a/templates/boot/grub/%SHORT_NAME%_default.cfg
+++ b/templates/boot/grub/%SHORT_NAME%_default.cfg
@@ -1,7 +1,7 @@
 menuentry "%GRML_NAME% - release %VERSION% (default)" {
     set gfxpayload=keep
     echo 'Loading kernel...'
-    linux   /boot/%SHORT_NAME%/vmlinuz apm=power-off boot=live live-media-path=/live/%GRML_NAME%/ bootid=%BOOTID% "${loopback}" "${kernelopts}" nomce net.ifnames=0 
+    linux   /boot/%SHORT_NAME%/vmlinuz apm=power-off boot=live live-media-path=/live/%GRML_NAME%/ bootid=%BOOTID% "${loopback}" ${kernelopts} nomce net.ifnames=0 
     echo 'Loading initrd...'
     initrd  /boot/%SHORT_NAME%/initrd.img
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by zeha** [comment]
Should the same be done for ${loopback}?

**Comment by michaellass** [comment]
I wondered this myself but I think for `${loopback}` the quotes might actually make sense.

It is set in [header.cfg](https://github.com/grml/grml-live/blob/eef40764c7ab2dbf8b2366f1dfb01b919be23091/templates/boot/grub/header.cfg#L32) and basically contains the path to the ISO if [set by grml-rescueboot](https://github.com/grml/grml-rescueboot/blob/f4f470fc86fa06a8fc57d406cf1156f60f171a9f/42_grml#L115). 

I would need to test this but theoretically the path to the ISO could contain spaces and then the quotes might be required. At least they should not hurt in this case because `${loopback}` only contains a single argument.

**Comment by michaellass** [comment]
I cannot think of a case where the quotes around `${loopback}` are required since grml-rescueboot fails to generate a valid grub entry when there is a space or a `"` in the path the to ISO. However, they do not hurt either. The linux cmdline contains the correct `findiso=...` argument.

I used the occasion to further test the `${kernelopts}` behavior. I configured the following in my `grub.cfg`:
```
kernelopts=" foo bar var1=something var2=\"something with spaces\" "
```

Without the change of this PR:
Bildschirmfoto von 2020-12-28 11-25-22

With the change:
Bildschirmfoto von 2020-12-28 11-26-44

The arguments are parsed mostly correctly. The only issue left is about the one argument that contains spaces itself and is therefore encapsulated in quotes within `${kernelopts}`. The content of `${kernelopts}` itself is correct and does not contain those backslashes. Here's a screenshot of the grub shell within the grml ISO:
Bildschirmfoto von 2020-12-28 11-53-22 (Kopie)

I assume that grub escapes the quotes on its own instead of pasting the `${kernelopts}` variable verbatim into the linux cmdline. I could not find a solution around this, so maybe this is a restriction that grml-rescueboot has to live with.

michaellass:
Is there anything left I can do here?

mika:
Not at all, just busy with other stuff. :)

Thanks for the PR and also poking us, I just took care of it!

PR #95, opened 2021-03-06
GRUB: drop loopback usage for balder10/FreeDOS.

Ionic:
Looks like a relic from the bad, old days.

It's certainly not needed any longer, since the initrd16 command can
load the file just fine and newer grub versions are not able to load
compressed files in such a manner, since the length (+2880) refers to
the uncompressed size. Grub will see that the file is smaller than that
and error out. If we were to give it the compressed size, decompression
would fail due to the buffer being too small.


mika:
Great catch, thanks for taking care and providing a PR! Merged and submitted! :)

PR #97, opened 2021-03-08
grml-live: isohybrid is the default, reflect that.

Ionic:
While isolinux's "manifold" mode was used by default a long time ago,
Grml switched to the isohybrid method in late 2011, mostly due to manifold
being incompatible with UEFI systems.

Documentation and configuration files weren't updated at that time,
though.

Do it now.


Ionic:
Note that [grml/grml.org/grml-live/index.html](https://github.com/grml/grml.org/blob/master/grml-live/index.html) will also have to be regenerated. I'll do that in a different PR there.

Ionic:
Needed doc fixup to not parse incorrectly.

mika:
Great, thanks for attention to detail and providing this PR! :+1:

PR #103, opened 2021-03-24
grml-live: clean up MIRROR_DIRECTORY if provided.

Ionic:
Won't actually cause harm, but a spurious empty directory in the resulting squashfs root image looks weird.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -180,7 +180,10 @@ umount_all() {
    fi
 
    umount "${CHROOT_OUTPUT}/grml-live/sources/" 2>/dev/null || /bin/true
-   [ -n "$MIRROR_DIRECTORY" ] && umount "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
+   if [ -n "${MIRROR_DIRECTORY}" ]; then
+     umount "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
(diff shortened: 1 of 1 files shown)


mika:
Thanks for attention to detail and the PR, merged! :tada:

Ionic:
I noticed that this causes some... unforeseen problems.

Quick walkthrough for how I used it and what I noticed:
  - `mkdir -p /mnt/hdd`
  - mount volume to `/mnt/hdd`
  - volume contains a directory called `local-packages` => `MIRROR_DIRECTORY=/mnt/hdd/local-packages`
  - `grml-live` creates `$chroot_path/mnt/hdd/local-packages` and bind-mounts `/mnt/hdd/local-packages` to that place
  - FAI is being executed and the chroot is created
  - `grml-live` unmounts `$chroot_path/mnt/hdd/local-packages` and runs `mkdir -p $chroot_path/mnt/hdd/local-packages` => all directories up to `$chroot_path` are being deleted because they are now empty
  - squashfs image is being created => squashfs image now won't contain the base directory `/mnt`

The issue here is that
  - `/mnt` is a base (FHS) directory and not owned by package
  - it's empty by default.

Other empty base directories would be affected by that as well.

I'll have to find a way to figure out which directory is a base directory and create an `rmdir -p` wrapper for that. That... seems to be difficult, though, since nothing properly owns the directories.

PR #106, opened 2022-01-24
grml-live.txt: output dir mount options; errors; manifold

akorn:
Explain that the output dir shouldn't be nodev, noexec or nosuid (some packages may fail to configure correctly with nosuid even if no problems are immediately apparent with the default builds).

Add information on how to find errors in FAI logs.

Improve the section on reverting manifold.


mika:
Thanks for taking care! :hugs:

PR #116, opened 2022-08-04
Exclude /var/lib/dkms/* from squashfs

akorn:
These files and directories are created by dkms while it builds modules on
demand and are only relevant during the module build/installation process.
The .ko files are then installed in /lib/modules; shipping /var/lib/dkms/*
inside the squashfs only takes up space without providing a tangible
benefit.

### DIFF
--- a/etc/grml/fai/config/package_config/ZFS
+++ b/etc/grml/fai/config/package_config/ZFS
@@ -1,0 +1,7 @@
+PACKAGES install
+
+build-essential
+zfs-dkms
+zfsutils-linux
+dwarves
+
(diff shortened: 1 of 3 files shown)


etc/grml/fai/config/package_config/ZFS
```diff
@@ -0,0 +1,7 @@
+PACKAGES install
+
+build-essential
+zfs-dkms
+zfsutils-linux
+dwarves
```
mika, line 6: Is the `dwarves` relevant for ZFS? If so, could we please add an according comment in the commit message?

nitpick: could we please sort the package list alphabetically? :)

mika (commented):
Thanks for your PR! :partying_face:

mika:
Many thanks, @akorn!

PR #117, opened 2022-08-21
Add support for building grml ISOs with zfs

akorn:
The `ZFS` FAI class will, via the `instsoft.ZFS` hook, install the packages
needed to build the zfs modules, build the zfs modules, then remove the
development packages again to avoid bloating the ISO. The ZFS
`package_config` now only installs `zfsutils-linux` so that the iso contains
the userspace zfs tools.

In my tests, the ZFS class now only increases the size of the ISO by about
3MB.

I wanted to rely on apt autoremove to get rid of automatically installed
packages, but for some reason as of 2.5.2 autoremove doesn't actually remove
some of the development packages (e.g. gcc-11) even though they're Priority
optional, marked as auto-installed, and not depended on by anything that
isn't auto-installed.

So, to avoid bloating the ISO with hundreds of MB of development packages,
the `instsoft.ZFS` script works out which packages it installs and removes
those explicitly, using potentially brittle parsing of `apt-get` output.
This has been tested to work and should do until a better solution is found.

### DIFF
--- a/etc/grml/fai/config/hooks/instsoft.ZFS
+++ b/etc/grml/fai/config/hooks/instsoft.ZFS
@@ -1,0 +1,81 @@
+#!/bin/bash
+# Filename:      ${GRML_FAI_CONFIG}/hooks/instsoft.ZFS
+# Purpose:       Build zfs modules in the chroot, then get rid of packages installed for this alone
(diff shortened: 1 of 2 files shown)


mika:
Finally had the chance to give it a try, LGTM - thanks! :partying_face:

Issue #109, opened 2022-03-23
Replace memtest86+ with pcmemtest?


mika:
memtest86+ doesn't support EFI boot, while https://github.com/martinwhitaker/pcmemtest claims to provide EFI support, though pcmtest is available in Debian testing + unstable only for now, though that might be worth giving it a try.

riesebie:
:+1:  :shipit:

evgeni:
Do we need to *replace*?

Can't we hide m86+ on UEFI and show both on BIOS boot?

Reason being twofold:
* Memory testing always was a bit religious, and we could try to offer whatever people prefer
* M86+ might get UEFI support in the future

evgeni:
And just like this, https://github.com/memtest86plus/memtest86plus happened, which is now "Memtest86+ v6.0 was based on PCMemTest, which was a fork and rewrite of the earlier Memtest86+ v5.01, which in turn was a fork of Memtest86." :exploding_head:

zeha:
memtest86 in experimental is already the pcmemtest variant. So its likely nothing needs to be done :-)

PR #120, opened 2022-11-10
Support Memtest86+ with UEFI

mika:
Memtest86+ up and until v5.01-3.1 didn't provide (U)EFI support. Starting with v6.00, the file memtest86+.bin no longer exists, but instead there are:

* /boot/memtest86+x32.bin
* /boot/memtest86+x32.efi
* /boot/memtest86+x64.bin
* /boot/memtest86+x64.efi

So let's start adding support for memtest86 with (U)EFI. Hopefully we don't need FAT16/8.3 compatibility any longer, so let's try using the files as-is.

Closes: https://github.com/grml/grml-live/issues/109


mika:
AFAICT this should work as needed, though I'd appreciate if someone else could also take a quick look at it :)

jkirk (approved):

PR #129, opened 2023-02-04
update grub and shim from bookwork

pasja:
After 581da7443c68c362a7677c905ab5c63eb23c5b73 and using the `debian` style secure boot grml will not start on machines with secure boot enabled, but fails with a `signature verification error`

After some investigation it turned out that we hit https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925550 with our boot binaries.

I have updated them from debian and managed to boot with them properly.

Checking those binaries should be a regular excersize in order to avoid situations like this.


pasja:
Due to the sensitive nature of this, I can fully accept if you do not merge this, but update them by yourself

mika:
@pasja thank your very much for bringing this to our attention and providing a PR! I just pushed the according updates, thanks! :partying_face:

PR #141, opened 2023-12-01
grml-docs/startpage.html + templates/boot/isolinux/f10: rework and update instructions

mika:
Re `grml-docs/startpage.html`:

* We don't ship xlinks2, so drop it from the browser list
* Iceweasel no longer exists
* There's no point in providing usage instructions for the command line browsers, other than how to exit them (if someone really ends up there without being used to w3/links)
* Replace all references to http with https
* Let's drop tldp.org / Linux Documentation Project (unsure whether this is relevant for anyone nowadays?), and instead link to Grml Wiki
* We moved our IRC from freenode to oftc, update accordingly
* While at it, misc rewordings and section adjustments, to reduce text on page

And in `templates/boot/isolinux/f10` I stumbled upon references to freenode


zeha (approved):

zeha:
Nice

jkirk (approved):

PR #142, opened 2023-12-07
Remove Grml release name from the boot options

jkirk:
The Grml release name is shown on top of the additional boot option menu and also in front of every boot option.
This limits the readable title length. We decided to remove it.

Changed "German Settings" to "Load German Keyboard Layout" because that is what it does (i.e. it runs `grml-lang de`).

Also changed the description of some entries as they were cut off, on the boot screen.
Closes: grml/grml#10


mika (approved):
Thanks for taking care! :heart:

PR #145, opened 2024-03-07
etc/fstab/GRMLBASE: use reserved IP following RFC5737

mika:
Usage of IP addresses like 1.2.3.4 is not recommended, as they might point to actually used and non-reserved IPs.

Quoting from RFC 5737 AKA "IPv4 Address Blocks Reserved for Documentation" (see https://datatracker.ietf.org/doc/html/rfc5737):

```
| The blocks 192.0.2.0/24 (TEST-NET-1), 198.51.100.0/24 (TEST-NET-2), and
| 203.0.113.0/24 (TEST-NET-3) are provided for use in documentation.
```

So use 192.0.2.42 instead.


zeha (approved):
Nice

jkirk (approved):

PR #146, opened 2024-03-10
TOR class to allow remote SSH access over deep networks

anarcat:
This class configures an onion service to access the SSH server over Tor. This is useful when you have a box behind NAT or some firewall that is broken or unknown, and you need to get a rescue shell on the host.

With this, you give an operator a thumb drive, who only needs to figure out how to boot into GRML, and then after a while you get a shell, pretty much regardless of where the box is.

This is not enabled by default, naturally, otherwise the secret key would leak in default GRML builds: this is solely designed to be run in an ad-hoc, one-time fashion. It also generates the SSH keys for the same reason: those are shown in the build logs and can be used to authenticate the remote host (a redundant measure to the onion service name, of course).

I also enable `DEFAULT_BOOT_OPTIONS=ssh` in my builds, but that hasn't been done here (although maybe it's possible to enable that in the class? to be investigated).

Finally, another shim is required here to inject a valid SSH public key in the image, so you can login over SSH. In my case, I have an extra CLASS that only has this one script which does:

    gpg --export-ssh-key anarcat@debian.org | tee -a $target/root/.ssh/authorized_keys

This could also be folded in the TOR class, but I'm not sure how to do variables yet, so that's not yet standardized.


anarcat:
This probably needs docs and maybe a built-in way to enable `ssh` and an authorized_keys by default, but I'd like to hear if it's useful from the team before going any further.

etc/grml/fai/config/scripts/TOR/10-gen_hs
```diff
@@ -0,0 +1,84 @@
...
+        process.terminate()
+
+with hostname_path.open() as hn:
+    print("Tor onion service hostnamename:", hn.read().strip())
+
+print("generating SSH keys in", target)
+run(["chroot", target, "ssh-keygen", "-A"], check=True)
+
+count = 0
+for path in [str(x) for x in (Path(target) / "etc/ssh").iterdir()]:
```
zeha, line 67: stringifying these Path objects here seems weird. why not use the Path instance attributes and methods instead of using string operations?

zeha:
> > we can't merge the `DEFAULT_BOOTOPTIONS` change as is, as it would apply to all builds. Maybe that should go into a readme for the class.
> 
> what would go in the README here actually? something like "you also need the `ssh` boot option in .live.local"?

yeah. and/or "you need to set DEFAULT_BOOTOPTIONS=ssh" when invoking grml-live.

Also I think the text from the PR description would also be useful in that readme.

docs/grml-live.8.adoc
```diff
@@ -378,6 +378,20 @@ official grml release
 * SOURCES: retrieve Debian source packages after installation. Files will be
 placed in the output directory under grml_sources.
 
+* TOR: generates a Tor onion service at build time. This makes the
+  booted system available even if behind NAT or some firewall that is
+  broken or unknown, and you need to get a rescue shell on the
+  host. Note that the Tor private keys are generated in the ISO, which
```
zeha, line 384:
```suggestion
  host. Note that the Tor private keys are baked in the ISO, which
```
?

zeha (approved):

mika (approved):

zeha (approved):

PR #149, opened 2024-05-25
Fix grml-live build from testing

pasja:
Both `eterm` and `hping3` has been removed from debian testing(trixie) and checking their tracker page I assume that they will not reenter soon.

* https://tracker.debian.org/pkg/hping3
* https://tracker.debian.org/pkg/eterm

In order to presevere the build-ability of testing let's remove those packages for now

### DIFF
--- a/etc/grml/fai/config/package_config/GRML_FULL
+++ b/etc/grml/fai/config/package_config/GRML_FULL
@@ -156,7 +156,6 @@ dnsutils
 ethstatus
 ethtool
 hostapd
-hping3
 ifenslave
 iftop
 ifupdown
--- a/etc/grml/fai/config/package_config/XORG
+++ b/etc/grml/fai/config/package_config/XORG
@@ -16,8 +16,6 @@ grml-x
 grml-desktop
 # osd_cat:
 xosd-bin
-# Esetroot:
-eterm
 # x-terminal:
 xterm
 # links2 provides a small X browser named xlinks2:


pasja:
@mika WDYT about this PR?

jkirk:
Thx @pasja for your PR. I've just looked into this:

The issue regarding `hping3` is worked on: https://salsa.debian.org/debian/hping3/-/merge_requests/4

And `eterm` has a pending patch: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1066342#10

Considering the fact that we do not have a schedule for the next version, we should not remove the packages.
@mika What do you think?

mika:
(Sorry for late reply.)

I very much appreciate your (@pasja) efforts to keep Debian/testing builds working! Removing packages from the package list has the risk of no longer being aware of possible show stoppers for our stable release though, as @jkirk also reported. So we should to be careful about this. For example `Esetroot` of the `eterm` package is used within our `grml-wallpaper` Zsh function, no big deal really, but still something we might overlook and then users might see unexpected behaviors. :)

Good news though: eterm is already back in Debian/testing as of yesterday (2024-06-06),

Bad news: the following packages are currently missing in Debian/testing:

* hwinfo: due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1069498, someone should look into this
* nocache: due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1069426, already forwarded to upstream (though I'm afraid upstream won't react), someone should look into this
* hping3: due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1066662, already fixed via https://salsa.debian.org/debian/hping3/-/merge_requests/4 but pending upload from maintainer (pinged him in the MR)

Especially hwinfo is relevant for us due its usage within grml-hwinfo (no hard dependency though). Would be great if we could bring those packages back to Debian/testing!

pasja:
Just checked all three linked debian bugs, and it seems that they have been closed and entered testing, so I close this PR.

PR #151, opened 2024-08-16
SW: drop reiser4progs (to be removed from Debian)

zeha:
Debian will drop the package, see
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078247


mika:
FYI: sorry for the `mika closed this now` and `mika deleted the drop-reiser4 branch` noise, this is triggered when I'm merging something else and `git push --mirror` runs from git.grml.org towards github.com :-/
Restored it for now

PR #152, opened 2024-08-20
Drop support for all boot methods other than isolinux

mika:
The "grub2" and "manifold" options for HYBRID_METHOD as well as the "grub" and "grub2" options for BOOT_METHOD are not used by any official Grml ISO, though they add considerable overhead and complexity in our code base.

I'd like to keep our code base simple and avoid feature creep, so let's deprecate BOOT_METHOD, and drop support for HYBRID_METHOD other than isolinux.

Note: not yet verified myself and there's also a `TODO` left (see code changes), but trying to gather early feedback :)


zeha:
haven't looked at the change yet, but does it preserve uefi boot with grub2?

etc/grml/grml-live.conf
```diff
@@ -152,12 +147,8 @@
 # boot the CD using normal el torito mode or copy it to USB device
 # *without* having to run grml2usb (like: 'dd if=grml.iso of=/dev/sdX')
 # - working both with the same ISO
-# Note that the manifold method is currently not be compatible with UEFI
-# setups.
 # HYBRID_METHOD='disable'   # do not create a hybrid ISO
```
zeha, line 150: is there a good reason to keep supporting `disable`?

zeha: lets consider this in a follow up issue

zeha (approved):

PR #153, opened 2024-09-06
Fix DHCP situation with grml-full unstable/sid

mika:
On our recent grml64-full sid/unstable daily ISOs we ended up with udhcpc (the busybox DHCP client implementation) instead of dhcpcd.

We build our daily ISOs with grml-live, using `-s sid -c DEBORPHAN,GRMLBASE,GRML_FULL,RELEASE,AMD64,IGNORE`. The class selection then gets mapped to FAI classes `DEBIAN_SID DEBORPHAN GRMLBASE GRML_FULL RELEASE AMD64 IGNORE`.

For handling the isc-dhcp-client/dhcpcd situation, we need to enable its according DEBIAN_* class (as we can handle this only for releases newer than bookworm, as cloud-init had a hard dependency on isc-dhcp-client up until and including bookworm, see https://bugs.debian.org/1051421 for details).

So in /etc/grml/fai/config/package_config/DEBIAN_UNSTABLE we already handle the isc-dhcp-client -> dhcpcd switch for unstable, but now that we mapped the `-s sid` to FAI class `DEBIAN_SID`, the `DEBIAN_UNSTABLE` configuration never got active.

If we'd add another DEBIAN_SID class file as duplicate of DEBIAN_UNSTABLE, this would mean further maintenance and debugging efforts. Instead let's drop the single one DEBIAN_SID configuration file we had so far, and map the `sid` codename to the DEBIAN_UNSTABLE class instead.

Thanks: @jkirk for assistance in tracking this down

### DIFF
--- a/debian/grml-live.maintscript
+++ b/debian/grml-live.maintscript
@@ -6,3 +6,4 @@ rm_conffile /etc/grml/fai/config/files/etc/systemd/system/serial-getty@ttyS0.ser
 rm_conffile /etc/grml/fai/config/scripts/GRMLBASE/36-cpufrequtils 0.33.0~
 rm_conffile /etc/grml/fai/config/scripts/GRMLBASE/40-deborphan 0.35.0~
 rm_conffile /etc/grml/fai/config/scripts/GRMLBASE/93-update-usbids 0.45.0~
+rm_conffile /etc/grml/fai/config/files/etc/apt/sources.list.d/debian.list/DEBIAN_SID 0.47.10~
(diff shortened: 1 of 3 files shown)


zeha (approved):

jkirk:
It's worth noting that cloud-init v24.1 moved to dhcpcd as the default dhcp client, see: https://github.com/canonical/cloud-init/releases/tag/24.1

I already filed a bug against the cloud-init Debian package, but still waiting for a Bug#.

jkirk (approved):

PR #158, opened 2024-10-04
Drop debian2hd boot option

zeha:
grml-autoconfig dropped this, as it is broken under systemd and nobody seems to be using it.

See https://github.com/grml/grml-autoconfig/pull/17


jkirk:
Relates to grml/grml#9.

mika (approved):

PR #169, opened 2024-11-04
SOURCES: move directory instead of bind-mounting it in

zeha:
Necessary for building in (unprivileged) docker, where bind mounts are not allowed.


mika (approved):
Untested/unverified but LGTM, thanks!

PR #170, opened 2024-11-05
SOURCES: run apt-get source parallelized

zeha:
Saves about 8min (~40%!) on a GRML_SMALL build.
Also calls apt-get less often, as we now ask dpkg for the source package names.


mika (approved):
Awesome, thanks!

PR #172, opened 2024-11-22
Drop vestigial genisoimage/mkisofs support

zeha:
Since 3c0d23bcd80c75ab4390d467f3bfceed42bce0c7 (from 2018!) grml-live hard-depended on `xorriso`. Remove leftover genisoimage/mkisofs support.

Also drop genisoimage from `GRML_FULL`, which already must include `xorriso` (for grml-live).

PR #174, opened 2024-11-22
Import CI build drivers

zeha:
These are our new build scripts for driving grml-live from CI. Not the nicest Python I've ever written, but it works.

This PR also contains #172 and #173 for now, as I needed them for testing in the same branch.


grml-live
```diff
@@ -637,32 +637,29 @@ extract_iso() {
 if [ -n "$EXTRACT_ISO_NAME" ]; then
   log "Unpacking ISO from ${EXTRACT_ISO_NAME}"
   einfo "Unpacking ISO from ${EXTRACT_ISO_NAME}"
-  local mountpoint=$(mktemp -d)
+  local tempdir=$(mktemp -d)
   local rc=0
-  mount -o loop "${EXTRACT_ISO_NAME}" "$mountpoint" ; rc=$?
+  mkdir -p "${tempdir}/live/"
+  osirrox -indev "${EXTRACT_ISO_NAME}" -extract live "${tempdir}/live/"
```
mika, line 643: missing the "; rc=$?" at the end to propagate the exit code for lines below?

mika, line 643: TIL about osirrox, nice! :+1:

grml-live
```diff
@@ -637,32 +637,29 @@ extract_iso() {
...
   if [ "$rc" != 0 ]; then
-    rmdir "$mountpoint"
-    log "mount failed"
-    eerror "mount failed"
+    rm -r "$tempdir"
```
mika, line 645: are we good/fine with `rm -r` (as in: no need for `rm -rf`)?

mika (commented):
Wow! :bow:

grml-live
zeha: indeed, fixed now!

grml-live
zeha: Was pondering that; have now added -f.

mika (approved):
Can't really comment on the python thingy, but everything LGTM!
:hugs:

Issue #160, opened 2024-10-19
linux-sysctl-defaults missing


zeha:
New package available in trixie and newer, necessary for ping to work.

PR #178, opened 2024-11-23
Install linux-sysctl-defaults on trixie and newer

zeha:
Brings the necessary sysctl settings for /bin/ping to work.

Closes grml/grml-live#160.


zeha:
17kB installed size

mika (approved):
Thx!

Issue #114, opened 2022-07-04
Re-evaluate initramfs compression?


mika:
We use `xz` compression for initramfs since ages™, as it provides a small initramfs file:

```
# time update-initramfs -k all -u
[...]
real    0m30.415s
user    2m40.279s
sys     0m3.720s
# ls -lah /boot/initrd.img-*
-rw-r--r-- 1 root root 38M Jul  4 17:13 /boot/initrd.img-5.18.0-2-amd64
# file /boot/initrd.img-*
/boot/initrd.img-5.18.0-2-amd64: XZ compressed data, checksum CRC32
```

But initramfs-tools uses zstd as default compression for initramfs starting with v0.141 (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=976054), and while disk usage increases, build time gets reduced quite significantly:

```
# time update-initramfs -k all -u
[...]
real    0m13.089s
user    0m14.102s
sys     0m3.094s
# ls -lah /boot/initrd.img-*
-rw-r--r-- 1 root root 47M Jul  4 17:12 /boot/initrd.img-5.18.0-2-amd64
# file /boot/initrd.img-*
/boot/initrd.img-5.18.0-2-amd64: Zstandard compressed data (v0.8+), Dictionary ID: None
```

We should figure out run-time issues (memory usage, time for uncompression), and if zstd performs better than xz, it might be worth switching to it as our new default?

zeha:
https://salsa.debian.org/kernel-team/initramfs-tools/-/merge_requests/37#note_211939
the decompression time improvement sounds compelling.

zeha:
Switchting to zstd enlarges the grml-small-arm64 ISO by 20MB.

zeha:
I say we do not change the current xz choice.

PR #180, opened 2024-11-23
WIP: drop COMPRESS=xz

zeha:
For evaluation of #114


mika:
So as it seems `xz` compression is still the preferred choice?

zeha:
Yeah I think it makes sense to keep XZ due to the space savings it provides. :-/

PR #181, opened 2024-11-23
Improve addons-absent code flow, drop syslinux on arm64

zeha:
Moves non-addon entries out of addons.cfg. With NO_ADDONS=1, addons.cfg is deleted, but those entries should still exist.

Then we set NO_ADDONS=1 early, if the addons directory is missing, instead of just complaining later.

With this, the same code path as for NO_ADDONS=1 is executed, when the addons directory is missing. This causes addons.cfg to be (correctly!) deleted.

Also add NO_ADDONS=1 state to grml-live configuration output, and always show the requested configuration, even if -F is given.

---

Additionally, this removes syslinux/isolinux/pxelinux from the arm64 builds. These tools are x86-only, and as such are only dead weight on arm64. Bonus effect: the resulting ISO is understood by macOS once again.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -30,7 +30,6 @@ GRML_LIVE_VERSION='***UNRELEASED***'
 # global variables
 PN=$(basename "$0")
 CMDLINE="$0 $*"
-ADDONS_LIST_FILE='/boot/isolinux/addons_list.cfg'
 # }}}
 
 # usage information {{{
@@ -315,6 +314,24 @@ copy_addon_file() {
   msg="Missing addon file: \"$1\""
   ewarn "$msg" ; eend 1
   log "copy_addon_file: $msg"
+  return 1
+}
+
+# Usage: copy_optional_file_logged /path/to/destination/filename /sources /src1 ...
+copy_optional_file_logged() {
+  local dest="$1"
+  shift
+  local source_root="$1"
+  shift
+  local source
+
+  for source in "$@" ; do
+    if [ -r "${source_root}${source}" ] ; then
+      log "Installing ${source} as ${dest}"
+      cp "${source_root}${source}" "${dest}"
+      return 0
+    fi
+  done
   return 1
 }
 
@@ -1077,8 +1094,7 @@ if [ -n "$BOOTSTRAP_ONLY" ] ; then
   log   "Skipping stage 'boot' as building with bootstrap only."
   ewarn "Skipping stage 'boot' as building with bootstrap only." ; eend 0
 else
-  # booting stuff:
-  mkdir -p "$BUILD_OUTPUT"/boot/isolinux
+  # Install boot code
   mkdir -p "$BUILD_OUTPUT"/boot/"${SHORT_NAME}"
 
   # this is a variable we're using for adjusting boot templates, not only in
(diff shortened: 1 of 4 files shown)


zeha:
Probably best to review the individual commits.

grml-live
```diff
@@ -507,6 +507,19 @@ if [ -n "$ZERO_LOGFILE" ] ; then
 fi
 # }}}
 
+# Automatically enable NO_ADDONS=1 if addons are not present {{{
+if [ "${NO_ADDONS:-}" != "1" -a ! -r "$TEMPLATE_DIRECTORY"/boot/addons ] ; then
```
mika, line 534: please use:

if [ "${NO_ADDONS:-}" != "1" ] && [  ! -r "$TEMPLATE_DIRECTORY"/boot/addons ] ; then

instead (also see shellcheck issue SC2166)

grml-live
```diff
@@ -541,6 +541,7 @@ echo "  main directory:    $OUTPUT"
 [ -n "$HYBRID_METHOD" ]       && echo "  Hybrid method:     $HYBRID_METHOD"
 [ -n "$SECURE_BOOT" ]         && echo "  Secure Boot:       $SECURE_BOOT"
 [ -n "$TEMPLATE_DIRECTORY" ]  && echo "  Template files:    $TEMPLATE_DIRECTORY"
+[ -n "$NO_ADDONS" ]           && echo "  Skipping boot addons."
```
mika, line 567: What's supposed to happen if one sets `NO_ADDONS=0`? :)

grml-live
```diff
@@ -481,6 +481,12 @@ specify it on the command line using the -c option."
 [ -n "$OUTPUT" ] || bailout 1 "Error: \$OUTPUT unset, please set it in $LIVE_CONF or
 specify it on the command line using the -o option."
 
+if [ "$ARCH" != i386 -a "$ARCH" != amd64 -a "$ARCH" != arm64 ] ; then
```
mika, line 501: please use `[ ... ] && [ ...] && [...]` style also here (also see shellcheck issue SC2166)

grml-live
```diff
@@ -1358,69 +1379,72 @@ else
     done
   done
 
-  # generate addon list
-  rm -f "${BUILD_OUTPUT}/${ADDONS_LIST_FILE}"
-  for name in "${BUILD_OUTPUT}"/boot/isolinux/addon_*.cfg ; do
-    include_name=$(basename "$name")
-    echo "include $include_name"  >> "${BUILD_OUTPUT}/${ADDONS_LIST_FILE}"
-  done
+  if [ "${ARCH}" = "amd64" -o "${ARCH}" = "i386" ] ; then
```
mika, line 1382: please let's use `if [ ... ] || [  ...] ` here also

grml-live
```diff
@@ -1723,81 +1752,98 @@ create_netbootpackage() {
...
   mkdir -p "$WORKING_DIR"
 
   cp "${CHROOT_OUTPUT}"/boot/vmlinuz-*    "$WORKING_DIR"/vmlinuz
   cp "${CHROOT_OUTPUT}"/boot/initrd.img-* "$WORKING_DIR"/initrd.img
-  cp "${CHROOT_OUTPUT}/${pxelinux_dir}/pxelinux.0" "${WORKING_DIR}/pxelinux.0"
 
-  if [ -r "${CHROOT_OUTPUT}"/usr/lib/syslinux/modules/bios/ldlinux.c32 ] ; then
-    cp "${CHROOT_OUTPUT}"/usr/lib/syslinux/modules/bios/ldlinux.c32 "${WORKING_DIR}"/
-  fi
+  if [ "$ARCH" = "amd64" -o "$ARCH" = "i386" ] ; then
```
mika, line 1763: samr

grml-live
```diff
@@ -1723,81 +1752,98 @@ create_netbootpackage() {
...
-      else
-        log   "No shimx64.efi for usage with PXE boot found (shim-signed not present?)"
-        ewarn "No shimx64.efi for usage with PXE boot found (shim-signed not present?)" ; eend 0
+      if [ "$ARCH" = "amd64" ] ; then
+        if ! copy_optional_file_logged "${WORKING_DIR}"/shim.efi "${CHROOT_OUTPUT}" \
+          /usr/lib/shim/shimx64.efi.signed \
+          /usr/lib/shim/shimx64.efi
+        then
+          log   "No shim{x64,aa64}.efi for usage with PXE boot found (shim-signed not present?)"
+          ewarn "No shim{x64,aa64}.efi for usage with PXE boot found (shim-signed not present?)" ; eend 0
```
mika, line 1819: maybe log this only as `No shimx64.efi for usage ...` as we don't look for `shimaa64.efi` here?

grml-live
```diff
@@ -1723,81 +1752,98 @@ create_netbootpackage() {
...
-      else
-        log   "No grubnetx64.efi for usage with PXE boot found (grub-efi-amd64-signed not present?)"
-        ewarn "No grubnetx64.efi for usage with PXE boot found (grub-efi-amd64-signed not present?)." ; eend 0
+      if [ "$ARCH" = "arm64" ] ; then
+        if ! copy_optional_file_logged "${WORKING_DIR}"/shim.efi "${CHROOT_OUTPUT}" \
+          /usr/lib/shim/shimaa64.efi.signed \
+          /usr/lib/shim/shimaa64.efi
+        then
+          log   "No shim{x64,aa64}.efi for usage with PXE boot found (shim-signed not present?)"
+          ewarn "No shim{x64,aa64}.efi for usage with PXE boot found (shim-signed not present?)" ; eend 0
```
mika, line 1837: same vice versa? :)

mika (commented):
Very nice! Only some minor nitpicks, otherwise LGTM, thx!

grml-live
zeha: yes!

zeha:
> Not sure whether you consider it ready already, but from what I can see it LGTM :) Thx!

Had to force-push once more after merging #194.

I dodged the NO_ADDONS=0 question by dropping that part.

mika (approved):
Thx, LGTM!

PR #185, opened 2024-11-24
grub: retitle grml boot menu entries

### LINKED ISSUE #175: Move version info in bootloader menus
#163 introduced new ISO names. We also introduce arm64.

We want the Bootloaders to show the version string in a new place instead of in each line, as the names are otherwise probably too long.

zeha:
The default menu entry now just reads "<grml-name> <version>", so for example "grml-small-arm64 2024.12". The additional options are in a menu entry just titled " -> Options".

To allow for removing the name/version for the options menu, grub.cfg now sorts the %SHORTNAME%_options.cfg always directly after the %SHORTNAME%_default.cfg.

Use unicode.pf2 to allow for Unicode characters in the Options and Addons menu entries.

Will need a trivial rebase after #181 is merged.

Closes #175.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -1306,8 +1306,7 @@ if [ "$ARCH" = i386 ] || [ "$ARCH" = amd64 ] || [ "$ARCH" = arm64 ] ; then
     fi
 
     # arch independent files
-    cp -a "${CHROOT_OUTPUT}"/usr/share/grub/ascii.pf2     "${BUILD_OUTPUT}"/boot/grub/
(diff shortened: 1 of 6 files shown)


mika (approved):
Hard to verify without actually testing it, but LGTM! Thaaaaaanks!

zeha:
> Hard to verify without actually testing it, but LGTM! Thaaaaaanks!

Right; a screenshot is in #186 :)

PR #189, opened 2024-11-26
GHA: setup shellcheck

zeha:
WIP obv.


zeha:
Running a local test build right now. I hope I didn't break anything, but not super confident.

evgeni:
FWIW: We've been using https://github.com/redhat-plumbers-in-action/differential-shellcheck which has the benefit of only failing on newly added violations, not on existing offenses

zeha:
> FWIW: We've been using https://github.com/redhat-plumbers-in-action/differential-shellcheck which has the benefit of only failing on newly added violations, not on existing offenses

Ah, very cool. I've "just" fixed all issues ;-)

evgeni:
Yeah, I'm never that motivated when working with shell ;)

zeha:
> Running a local test build right now. I hope I didn't break anything, but not super confident.

This found an issue or two which I've fixed.

scripts/generate_netboot_package.sh
```diff
@@ -43,7 +43,7 @@ cat > "${WORKING_DIR}/pxelinux.cfg/default" << EOF
 default grml
 label grml
   kernel vmlinuz
-  append initrd=initrd.img root=/dev/nfs rw nfsroot=192.168.0.1:/live/image boot=live apm=power-off quiet nomce noprompt noeject vga=791 net.ifnames=0 
+  append initrd=initrd.img root=/dev/nfs rw nfsroot=192.168.0.1:/live/image boot=live apm=power-off quiet nomce noprompt noeject vga=791 net.ifnames=0
```
mika, line 46: please don't drop the trailing whitespace char here (it's relevant when e.g. someone wants to append something like `ssh=foobar`)

etc/grml/fai/config/hooks/instsoft.GRMLBASE
```diff
@@ -23,17 +26,17 @@ if [ "$FAI_ACTION" = "softupdate" ] ; then
 
    # /etc/resolv.conf is usually a symlink, pointing out of the chroot.
    # Make it a file with known contents.
-   rm -f "${target}"/etc/resolv.conf
+   rm -f "$target"/etc/resolv.conf
```
mika, line 29: nit: we have a strange mixture of `"${target}"` vs `"$target"` now :p
(I'm aware that this is inconsistent over all the code base, nowadays I prefer to use `"$f"` and `"${foo}"` (so no `${}` for single char variables, whereas `${}` for >=2 chars)

etc/grml/fai/config/hooks/instsoft.ZFS
```diff
@@ -29,9 +32,10 @@ echo "$0: Installing latest kernel and its headers, as well as build-essential."
 # keeping track of what gets installed. This is an ugly hack and should not
 # be needed, but without it the resulting ISO is hundreds of megabytes
 # larger. I hope this kludge can go away eventually.
-extra_packages=($($ROOTCMD apt-get --assume-no --download-only --mark-auto -u install \
+mapfile -t extra_packages <($ROOTCMD \
```
mika, line 35: this doesn't work, AFAICT, needs to be:

`mapfile -t extra_packages < <($ROOTCMD  ...`

etc/grml/fai/config/hooks/instsoft.ZFS
```diff
@@ -54,7 +58,11 @@ else
 fi
 
 echo "$0: Installing zfs-dkms itself."
-extra_packages=(${extra_packages[@]} $($ROOTCMD apt-get --assume-no --download-only --mark-auto -u install zfs-dkms | sed '0,/The following NEW packages will be installed/d;/^[^ ]/,$d'))
+mapfile -t zfs_packages <($ROOTCMD \
```
mika, line 61: same

mika (commented):
Sorry for delay, had to find some proper time to go through all of this :)
Thanks a bunch for working on this, only some small nitpicks and two bugs, other than that LGTM!

etc/grml/fai/config/hooks/instsoft.GRMLBASE
zeha: Yeah, I tried to aim for some local consistency... at some point I gave up. We need to make a call and then update all places in one go.

mika: *nod* :)

etc/grml/fai/config/hooks/instsoft.ZFS
zeha: Fixed as above

etc/grml/fai/config/hooks/instsoft.GRMLBASE
zeha: Fixed at least instsoft.GRMLBASE to be in one style overall.

scripts/generate_netboot_package.sh
zeha: Ack, that was an overzealous text editor and an oversight on my part.

mika (approved):
Thanks, LGTM! :partying_face:

PR #191, opened 2024-11-26
scripts: remove obsolete code

zeha:
Remove various bits that seem obsolete since at least Debian buster.


etc/grml/fai/config/scripts/GRMLBASE/37-portmap
```diff
@@ -1,29 +0,0 @@
-#!/bin/bash
-# Filename:      ${GRML_FAI_CONFIG}/config/scripts/GRMLBASE/37-portmap
```
mika, line 2: When removing this file (being `/etc/grml/fai//config/scripts/GRMLBASE/37-portmap`) from the package, we also need to handle this via `rm_conffile` in `debian/grml-live.maintscript` so the file gets purged on package upgrades. (FTR, happy to rethink the way we ship the /etc/grml/fai` configuration files overall!)

mika (commented):
Just a comment regarding config file removal, rest LGTM, thanks for you attention to detail!

etc/grml/fai/config/scripts/GRMLBASE/37-portmap
zeha: Thanks for catching that, fixed now.

(Can think about the future after this release, I'd say.)

mika (approved):
LGTM!

PR #192, opened 2024-11-26
GRMLBASE/94-update-smart-drivedb: drop

zeha:
smartmontools >=6.4+svn4214-1 is satisfied since before Debian buster, and this script then does nothing.


zeha:
Didn't see this earlier, sorry!

PR #194, opened 2024-11-26
grml-live: always print requested configuration

zeha:
Skip the prompt with -F, but always print the configuration.


zeha:
(This includes all commits from #189 as it would just conflict...)

zeha:
Extracted from #181

PR #199, opened 2024-12-10
grml-live: remove GRML_SMALL exclude of vmlinuz, initrd.img

zeha:
The exclusion does not work today. /boot/ inside the squashfs always has vmlinuz* and initrd.img* files. IMO if the exclude would work, it would break build-only mode later.

A resulting grml-small build-only ISO does not change in size.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -1519,11 +1519,6 @@ else
       SQUASHFS_OPTIONS="$SQUASHFS_OPTIONS -ef $SQUASHFS_EXCLUDES_FILE -wildcards"
    fi
 
-   # get rid of unnecessary files when building grml-small for final release:
-   if hasclass GRML_SMALL ; then
-      SQUASHFS_OPTIONS="$SQUASHFS_OPTIONS -e initrd.img* vmlinuz*"
-   fi
-
    # log stuff
    SQUASHFS_STDERR="$(mktemp -t grml-live.XXXXXX)"
 


grml-live
```diff
@@ -1519,11 +1519,6 @@ else
       SQUASHFS_OPTIONS="$SQUASHFS_OPTIONS -ef $SQUASHFS_EXCLUDES_FILE -wildcards"
    fi
 
-   # get rid of unnecessary files when building grml-small for final release:
-   if hasclass GRML_SMALL ; then
-      SQUASHFS_OPTIONS="$SQUASHFS_OPTIONS -e initrd.img* vmlinuz*"
-   fi
-
```
mika, line 1526: Aha interesting, `/usr/share/doc/squashfs-tools/changelog.gz` says:

```
1.3r3   18 JAN 2004     Third release of 1.3, this adds a new mksquashfs option,
                        some bug fixes, and extra patches for new kernels

        1. New mksquashfs -ef exclude option.  This option reads the exclude                                                                                                                                                                                                  
           dirs/files from an exclude file, one exclude dir/file per line.  This
           avoids the command line size limit when using the -e exclude option,
```

But the `-e` option of mksquashfs isn't even listed anymore in `mksquashfs --help`. :-/

mika (approved):
Yeah let's apply this, we can look into this again if we're trying to optimize for disk space reasons, though not shipping the files asks for more troubles in our workflow for now.

Thanks!

grml-live
zeha: and this dropped -e? oh well

PR #201, opened 2024-12-10
GHA: run grml-live in test-build workflow

zeha:
Add an invocation of grml-live to the test-build GitHub Actions workflow. The results are collected as artifacts.

I consider this an initial start of having some form of end-to-end test for grml-live.
What I don't want here is to "test" Debian itself. For this reason, the test setup creates a minimal flavour/class. But at least GRMLBASE must work.


mika (approved):
Awesome! :heart:

Issue #200, opened 2024-12-10
Consider masking ldconfig.service


zeha:
ldconfig.service seems to use substantial time on boot. Given it's a live image, I don't think we need it.

Consider masking ldconfig.service somewhere in GRMLBASE scripts?

mika:
That's a great idea, ack thanks!

PR #203, opened 2024-12-10
GRMLBASE/15-initsetup: mask ldconfig.service

zeha:
ldconfig.service is mostly useful if the OS image can change its composition during boot. This is not the case for us.

As ldconfig.service can be quite slow, lets mask it.

Closes: grml/grml-live#200


mika (approved):

zeha:
I don't have hard numbers on how much time this saves, but it seems better to me.

For verification purposes, I've manually run `ldconfig -X` on an ISO with this change, and the resulting ld.so.cache is bit-identical to the one already in the squashfs.

Issue #90, opened 2020-06-24
localepurge: wrong config, maybe other issues


jkirk:
Just noticed that `grml-full` generates 83 locales which consumes nearly 42MB:

```
root@grml ~ # grml-version 
grml64-full_testing build3165 Release Codename autobuild-build3165 [2020-06-24]
root@grml ~ # grep -v "^#" /etc/locale.gen | wc -l
83
root@grml ~ # l /usr/lib/locale 
total 40508
drwxr-xr-x 3 root root      236 Jun 24 04:04 C.UTF-8
-rw-r--r-- 1 root root 41479488 Jun 24 04:22 locale-archive
```

Removing all locales except `en_US.UTF-8` drops the disk usage to about 3MB:

```
root@grml ~ # echo "en_US.UTF-8 UTF-8" > /etc/locale.gen
root@grml ~ # locale-gen 
Generating locales (this might take a while)...
  en_US.UTF-8... done
Generation complete.
root@grml ~ # l /usr/lib/locale     
total 2892
drwxr-xr-x 3 root root     236 Jun 24 04:04 C.UTF-8
-rw-r--r-- 1 root root 3035216 Jun 24 18:58 locale-archive
```

I believe GRMLBASE and GRMLFULL (in https://github.com/grml/grml-live/tree/master/etc/grml/fai/config/files/etc/locale.gen) should only provide `en_US.UTF-8` as we use localepurge which is configured to remove all locales except `en_US.UTF-8` and `en`.

zeha:
f3b335ea5c94d5cf24ec6b9184106b298849f311 is supposed to install a locale.gen with a restricted locale list.

zeha:
In my testing of localepurge it defaults to keeping the locales from /etc/locale.gen. Not sure what exactly you've observed ...

zeha:
`/etc/locale.nopurge` on grml-full lists only `en` and `en_US.UTF-8`. This kinda implies our locale setup on grml-full is broken.

Given the intention of having quite some locales installed, at least localepurge should be configured correctly.

zeha:
<img width="716" alt="Image" src="https://github.com/user-attachments/assets/f1a74fd7-777d-43b0-bf2c-da2e50a116bc" />

I think we're good!

PR #207, opened 2024-12-11
SW: GRML_FULL: stop precreating locales

zeha:
Given GRML_FULL also includes localepurge, which we expect to purge everything except en_US.UTF-8, creating all these locales seems pointless.

The FAQ also documents that only en_US.UTF-8 is supported.

Tentative fix for #90.


mika (approved):
Thanks for looking into this!

PR #208, opened 2024-12-13
Initialize zsh caches so first startup on Live CD is faster

zeha:


etc/grml/fai/config/scripts/RELEASE/98-clean-chroot
```diff
@@ -45,5 +45,14 @@ $ROOTCMD chown "${USERNAME}:${USERNAME}" "/home/${USERNAME}"
...
 $ROOTCMD su -s /bin/sh "${USERNAME}" -c "rsync -Hav /etc/skel/ /home/${USERNAME}/"
 
+
+# Initialize zsh caches so first startup on Live CD is faster.
+echo "exit" | $ROOTCMD zsh -l -s -i
+echo "exit" | $ROOTCMD setpriv --reuid="${USERNAME}" --regid="${USERNAME}" --clear-groups --reset-env zsh -l -s -i
+ls -la "$target"/root/
+ls -la "$target"/home/"${USERNAME}"
+ls -la "$target"/root/.zcompdump
+ls -la "$target"/home/"${USERNAME}"/.zcompdump
```
mika, line 55: should we keep all of the `ls -la`?

zeha: ah maybe only the ones with .zcompdump. it also serves as an assert the cache got written.

or should i remove all of them?

mika: ah makes sense! hm, maybe let's keep all of them, though maybe also report upfront which files/directories we're listing to make it more obvious what's going on? :thinking:

zeha: pushed, please check :)

mika (approved):
:rocket:

PR #212, opened 2024-12-17
SW: Install openssh-client-ssh1 for DSA-keyed servers

zeha:
The current OpenSSH version has dropped support for DSA keys. Users might find them in a situation where they still need DSA keys when they're running our Live CD, so lets provide openssh-client-ssh1 too. At least for a while, until either openssh-client-ssh1 becomes unsupportable or enough time has passed.

Also drop "ssh" from GRML_FULL, as GRMLBASE already has openssh-client and openssh-server.

Adds about 1.5MB installed size.

### DIFF
--- a/etc/grml/fai/config/package_config/GRML_FULL
+++ b/etc/grml/fai/config/package_config/GRML_FULL
@@ -178,6 +178,7 @@ netcat-openbsd
 nethogs
 netsniff-ng
 nmap
+openssh-client-ssh1
 radvd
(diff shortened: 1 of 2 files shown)


mika (commented):
I love the idea, though don't see an openssh-client1 package in Debian, am I missing something? :)

zeha:
fixed ^ 2

mika (approved):

PR #214, opened 2024-12-18
Do not add /dev to squashfs

zeha:
The resulting device nodes are problematic when unpacking inside unprivileged podman containers, and the livecd does not need them to work.


mika (approved):

zeha:
Manually checked the resulting ISO from GHA CI, looking good.

PR #217, opened 2024-12-18
Ensure files are copied with fcopy into chroot with expected permissions

mika:
When running grml-live from within the git repository that was cloned by a user other than root, the files installed by FAI's fcopy have wrong/unexpected permissions.

For example:

```
| # stat /home/mika/build/grml-live-2024-12/grml_chroot/etc/rsyslog.conf
|   File: /home/mika/build/grml-live-2024-12/grml_chroot/etc/rsyslog.conf
|   Size: 1106            Blocks: 8          IO Block: 4096   regular file
| Device: 253,2   Inode: 4472261     Links: 1
| Access: (0644/-rw-r--r--)  Uid: ( 1000/    mika)   Gid: ( 1000/    mika)
| Access: 2024-12-18 15:39:59.477471695 +0100
| Modify: 2023-08-07 13:32:24.000000000 +0200
| Change: 2024-12-10 17:02:16.177262515 +0100
|  Birth: 2024-12-10 17:02:16.177262515 +0100
```

fcopy provides the following options:

```
| % fcopy -h
| fcopy, copy files using classes.
| [...]
|    -m user,group,mode   Set user, group and mode for copied files.
|    -M                   Same as -m root,root,0644
| [...]
```

Execute all fcopy command lines with `-M`, except for the ones that need execute permission (being /usr/share/initramfs-tools/scripts/init-top/grml and /etc/initramfs-tools/hooks/000-udev-shutup) for which we use the `-m` option accordingly then.

### DIFF
--- a/etc/grml/fai/config/scripts/GRMLBASE/20-sudo
+++ b/etc/grml/fai/config/scripts/GRMLBASE/20-sudo
@@ -15,7 +15,7 @@ target=${target:?}
 # shellcheck source=/dev/null
 . "$GRML_LIVE_CONFIG"
 
-fcopy -v /etc/sudoers
+fcopy -m root,root,0440 -v /etc/sudoers
 sed -i "s/\$USERNAME/$USERNAME/" "$target"/etc/sudoers
 chmod 440 "$target"/etc/sudoers
 chown 0:0 "$target"/etc/sudoers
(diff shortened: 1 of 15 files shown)


etc/grml/fai/config/scripts/GRMLBASE/20-sudo
```diff
@@ -15,7 +15,7 @@ target=${target:?}
 # shellcheck source=/dev/null
 . "$GRML_LIVE_CONFIG"
 
-fcopy -v /etc/sudoers
+fcopy -M -v /etc/sudoers
```
zeha, line 18: so technically this should be 0440, but good enough.

mika: Good catch, done in new revision!

zeha (approved):

PR #218, opened 2024-12-18
Rework wallpaper handling

zeha:
Previously we used fai to copy the base image for the desktop wallpaper, but used the font directly from the host. Unify this situation by using both from the host. To find the font and the wallpaper, put them both into TEMPLATE_DIRECTORY, and include this variable in CONFIGDUMP.

FONTFILE and GRML_BG can no longer be overridden, but must be customized in templates/wallpaper instead.


mika (approved):

zeha:
Force pushed to rebase on top of your changes.

Issue #209, opened 2024-12-14
EFI / Grub: error message in sub menu entry


zeha:
<img width="912" alt="Image" src="https://github.com/user-attachments/assets/f84c8419-654d-498e-9da6-10849c6f8c10" />

After pressing a key it boots, but this shouldn't happen.
amd64 is fine.

I'm also wondering if the console=... option is even correct for arm64.

zeha:
The error message from grub happens for all options in the Options menu.

jkirk:
The error message is shown when selecting _any_ sub menu entry under "Options".

Taking over from my duplicate issue grml/grml#214:

**Describe the bug**
Booting Grml in EFI mode brings the new Grub  boot theme (see: grml/grml-live@b0b979e).
When selecting any sub menu entry under "Options" the following error is shown:

> "error: null src bitmap in grub_video_bitmap_create_scaled."

There is a "Press any key to continue..." prompt, but Grub continues after a few seconds.

**Additional context**

I found the following hints:

* [#922161](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=922161) grub2: null src bitmap error while trying to drop background image
* https://superuser.com/questions/1846090/in-grub2-what-does-error-null-src-bitmap-in-grub-video-bitmap-create-scaled

(Changing the title of this issue and closing grml/grml#214.)

PR #219, opened 2024-12-18
GRUB theme: replace desktop-color with desktop-image

jkirk:
In b0b979ef the GRUB theme was replaced.

When selecting any sub menu entry under "Options" the following error is shown:

  | error: null src bitmap in grub_video_bitmap_create_scaled.
  | Loading kernel...
  | Loading initrd...
  |
  | Press any key to continue...

We already ran into this. It was fixed by replacing the desktop-color with a small black.png as desktop-image, see: 8a09c98f and grml/grml#16

Closes: grml/grml-live#209


mika (approved):

zeha:
thanks!

PR #226, opened 2024-12-18
templates: drop deprecated boot addons from isolinux/grub configs and grml-cheatcodes.txt

mika:
The following addons are no longer supported:

* Grub1
* All-in-One-Image
* FreeDOS

And GPXE is actually iPXE, so update its description accordingly.

Thanks to @jkirk for looking into this!

See https://github.com/grml/grml/issues/211


zeha (approved):

jkirk (approved):
LGTM

PR #245, opened 2024-12-20
49-sshd: create dropin file instead of using sed

zeha:
Should be more stable, and easier to understand.

### DIFF
--- a/etc/grml/fai/config/scripts/GRMLBASE/49-sshd
+++ b/etc/grml/fai/config/scripts/GRMLBASE/49-sshd
@@ -17,28 +17,12 @@ if ! [ -r "${target}/etc/ssh/sshd_config" ] ; then
   exit 0
 fi
 
-# make sure root login works, it's set to "without-password" since openssh-server v1:6.6p1-1
-# and defaults to "prohibit-password" since openssh-server v1:7.1p1-1
-if grep -q '^PermitRootLogin ' "${target}/etc/ssh/sshd_config" ; then
-  # make sure we don't modify our own disabled snippet once again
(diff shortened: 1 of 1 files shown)


etc/grml/fai/config/scripts/GRMLBASE/49-sshd
```diff
@@ -17,28 +17,13 @@ if ! [ -r "${target}/etc/ssh/sshd_config" ] ; then
...
-if grep -q '^UseDNS ' "${target}/etc/ssh/sshd_config" ; then
-  # make sure we don't modify our own disabled snippet once again
-  if ! grep -q 'UseDNS .*disabled via grml-live' "${target}/etc/ssh/sshd_config" ; then
-    sed -i "s/^\(UseDNS .*\)/# \1 # disabled via grml-live\nUseDNS no/" "${target}/etc/ssh/sshd_config"
-  fi
-else
-  echo "# Added via grml-live script:" >> "${target}/etc/ssh/sshd_config"
-  echo "UseDNS no" >> "${target}/etc/ssh/sshd_config"
-fi
+cat > "${target}"/etc/ssh/sshd_config.d/grml-live.conf <<EOT
```
mika, line 20: I like it, thx! Probably we should check whether the grml-live marker is already present in the file, otherwise we'd append the same config during every grml-live invocation (e.g. when running `-b` + `-u`)?

zeha: With ">" it just gets overwritten, no?

mika: Oh it's even a custom file, nice - nevermind! :)

mika (approved):

PR #247, opened 2024-12-20
ttys: set WorkingDirectory to $HOME

zeha:
Let systemd set the current working directory for tty1-tty4 to the users (either root or $USERNAME) home directory. Then we can remove hacks in zshenv/zshrc.


mika (approved):
This is great idea, thx!

PR #248, opened 2024-12-22
console-setup: create setupcon cache

zeha:
Previously console-setup would on startup find and copy the keymap and console font into /etc/console-setup.

This moves this caching step to grml-live execution time, and thus out of the (first) boot.

Files affected:

* /etc/console-setup/cached_UTF-8_del.kmap.gz
* /etc/console-setup/cached_Uni3-Terminus16.psf.gz
* /etc/console-setup/cached_setup_font.sh
* /etc/console-setup/cached_setup_keyboard.sh
* /etc/console-setup/cached_setup_terminal.sh


mika (approved):
:rocket: :+1:

PR #249, opened 2024-12-23
Switch from ftp.debian.org to deb.debian.org

zeha:
Both point to the same CDN nowadays, but the deb.debian.org name is more popular with automatic caching redirects et al.

While at it, drop the apt sources.list file for `DEBIAN_ETCH`.


mika (approved):
Good idea!

Issue #246, opened 2024-12-20
Reconsider fai conffiles


zeha:
Currently grml-live installs all the FAI config into /etc/grml/fai. They end up being dpkg conffiles.
This is annoying to maintain.

Consider if we can move these files to /usr/share.

Additionally, consider how people are actually using grml-live; do they/we need the .deb at all?

mika:
ACK, let's add a sanity check to make sure that folks how don't use `/usr/share/grml-live/...` or alike but still point to `/etc/grml/fai` get an according failure, then announce the change and migration steps via `debian/NEWS`

PR #250, opened 2024-12-24
Move config into /usr/share/grml-live/config

zeha:
This stops having /etc/grml/fai/config as dpkg conffiles.
Users will need to update `GRML_FAI_CONFIG` (or `-D`) if they set it.

Warns if `/etc/grml/fai/config` still exists and `GRML_FAI_CONFIG` is unset; also warns if `GRML_FAI_CONFIG/config` exists, as `GRML_FAI_CONFIG` now should point directly to the config directory, and not its parent.

fai.conf is now treated as an internal detail, like nfsroot.conf. It cannot be user-modified anymore.

`GRML_FAI_CONFIG` should now point to what /etc/grml/fai/config was before. This means users likely need to append `/config` to their existing `GRML_FAI_CONFIG` values.

Closes: #246


zeha:
Contains commits from #249 as they would otherwise conflict.

grml-live
```diff
@@ -748,7 +764,16 @@ if [[ -n "${BOOT_METHOD:-}" ]] && [[ "${BOOT_METHOD}" != "isolinux" ]] ; then
...
 
-# generate nfsroot configuration for FAI on the fly
+# generate FAI configuration on the fly
+FAI_CONF_DIR=$(mktemp -d)
+
+echo "# This is an automatically generated file by grml-live.
+# Do NOT edit this file, your changes will be lost.
+LOGUSER=
+FAI_LOGPROTO=ssh
+MNTPOINT=/media/mirror
```
mika, line 774: I think also both `FAI_LOGPROTO` and `MNTPOINT` could be unset by default?

I recall that I had the need for:

```
ROOTCMD="chroot $FAI_ROOT"
```

in fai.conf as a workaround for https://bugs.debian.org/989547, though this doesn't seem to be relevant any longer also for our container environments, right? :thinking:

zeha: We're not using it in the daily/release build jobs, so I don't think its relevant anymore?

FAI_LOGPROTO/MNTPOINT seemed a bit useless to me, but I haven't dug in. I've just kept the settings as is for now.

zeha: FAI_LOGPROTO/MNTPOINT: seems fine to remove them.

zeha: Removed them in the next commit.

jkirk:
I just created an ISO with the (adjusted) `run-local-test-build` / podman build environment from [grml/build-daily@9210ba5](https://gitlab.grml.org/grml/build-daily/-/commit/9210ba5839c41cff5e5c9cc166949261a7166821). LGTM.

debian/NEWS
```diff
@@ -1,3 +1,37 @@
+grml-live (0.50.2) unstable; urgency=low
+
+  The location of the FAI configuration data has changed, and
+  all users who do not use the default configuration data must
+  update their configuration.
+
+  In the Debian package, the configuration data files were
+  previously installed into /etc/grml/fai/config. To avoid
+  continous upgrading headaches, these files are now installed
```
mika, line 9: nit: s/continous/continuous/

debian/NEWS
```diff
@@ -1,3 +1,37 @@
...
+  into /usr/share/grml-live/config instead.
+  If you intend to modify them, I strongly recommend either
```
mika, line 11: nit: s/I/it is/

debian/NEWS
```diff
@@ -1,3 +1,37 @@
+grml-live (0.50.2) unstable; urgency=low
```
mika, line 1: maybe let's make this a 0.51.0 release, given that it's a major change?
(if so, debian/grml-live.maintscript needs to be adapted accordingly then also)

mika (commented):
LGTM overall, I'd just vote for raising the grml-live version and two minor nitpicks :)
Thx, looking forward to this!

debian/NEWS
ft: "…recommended" in that case too.

debian/NEWS
zeha: Bumped to 0.51.0

debian/NEWS
zeha: Applied both changes

debian/NEWS
zeha: Thx, applied

mika (approved):
Thx! :rocket:

PR #254, opened 2024-12-24
Remove vmwgfx enable_fbdev option

zeha:
Linux upstream removed it in https://patchwork.freedesktop.org/patch/508110/ which landed in Linux 6.2.

This was backported, as Debian's 6.1.0-28-amd64 no longer has the option, while 6.1.0-15-amd64 still had it. Linux 6.1 stable commit: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=67adcfae2eb177303ceba1fbcb0935909c6fecfa

Debian did enable it by default in 5.16, so enable_fbdev=1 was not necessary anymore on Debian bookworm. https://salsa.debian.org/kernel-team/linux/-/commit/2b167004cd95c08295f855fd5cacd05d158c5d46


mika (approved):
:heart:

Issue #128, opened 2023-02-02
Unify copy_addon_file behavior


mika:
Derived from grml/grml#178, we've a mixture of memtest binaries on our release ISOs, caused by files not used from the chroot but from the build system:

```
[...]
Installing boot addons.
Copying ipxe.lkrn from system
Copying ipxe.efi from system
Copying pci.ids from chroot
Copying memtest86+.bin from system
Copying memtest86+x32.bin from chroot
Copying memtest86+x32.efi from chroot
Copying memtest86+x64.bin from chroot
Copying memtest86+x64.efi from chroot
[...]
```

Being independent from specific build host is preferred and we should investigate if it's possible to get there.
As a start look into installing ipxe by default in our Grml ISOs (full vs. small might be annoying though as it's ~4MB of disk space:-/). Furthermore try to resolve the memtest confusion of ending up with 5 binaries on a ISO.

PR #256, opened 2024-12-26
Replace copy_addon_file

zeha:
Closes: #128

This removes copy_addon_file and uses copy_file_logged instead. There is no way of copying files from the host, or from ${TEMPLATE_DIRECTORY}/compat anymore. If necessary, please build with a newer Debian suite.

Hides ipxe, netboot.xyz from grub boot menu if the files are missing.
memtest86+ is not hidden, but instead the package is added to GRML_SMALL instead.

ipxe is not installed into GRML_SMALL, so it will be non-functional in isolinux on grml-small and absent from the grub boot menu.


mika:
> Errors that need addressing:
> 
> ```
> cp: cannot create regular file '/tmp/tmp2o3jma1w/grml_cd/boot/addons/pci.ids': No such file or directory
> ```

Just for my understanding before I'm starting the review: is this already handled or does it need to be fixed yet?

zeha:
> > Errors that need addressing:
> > ```
> > cp: cannot create regular file '/tmp/tmp2o3jma1w/grml_cd/boot/addons/pci.ids': No such file or directory
> > ```
> 
> Just for my understanding before I'm starting the review: is this already handled or does it need to be fixed yet?

I fixed that; however maybe wait with the review until the other PRs are merged.

zeha:
Now it should be good to review! :)

mika (approved):
Didn't check whether all boot menu entries still work also under e.g. grml2usb, but the code changes LGTM and it all makes perfect sense to go into that direction, thanks for working on this! :rocket: :heart:

PR #258, opened 2024-12-28
Stop installing memdisk

zeha:
Unnecessary since 530ec9ff0ac712bc4dab45ea820725d863154c59.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -1289,8 +1289,6 @@ else
       for file in chain.c32 hdt.c32 mboot.c32 menu.c32; do
         copy_addon_file "${file}" "${syslinux_modules_dir}" addons
       done
-
-      copy_addon_file memdisk /usr/lib/syslinux addons
     fi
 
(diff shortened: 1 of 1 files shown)


mika (commented):
Just wondering, do we have any "interface" or alike so folks still relying on `memdisk` in their templates get the according syslinux memdisk file installed as expected? :thinking:

zeha:
> Just wondering, do we have any "interface" or alike so folks still relying on `memdisk` in their templates get the according syslinux memdisk file installed as expected? 🤔

Not really. If its required, I guess they can copy the file into their addons folder?

mika:
> > Just wondering, do we have any "interface" or alike so folks still relying on `memdisk` in their templates get the according syslinux memdisk file installed as expected? 🤔
> 
> Not really. If its required, I guess they can copy the file into their addons folder?

ACK, let's find out! :)

mika (approved):

PR #259, opened 2024-12-28
GRMLBASE/15-initsetup: init journald catalog cache at build time

zeha:
systemd-journal-catalog-update.service updates the journald catalog cache. Create this cache at grml-live build time, and not on boot.


mika (approved):
Nice find! :+1:

PR #260, opened 2024-12-28
GRMLBASE/21-usersetup: remove grep of group file

zeha:
adduser is supposed to be idempotent; with -q it also does not log a warning when the user is already in the given group.

### DIFF
--- a/config/scripts/GRMLBASE/21-usersetup
+++ b/config/scripts/GRMLBASE/21-usersetup
@@ -31,7 +31,7 @@ fi
 add_user_to_group() {
    [ -n "$1" ] || return 1
    if grep -q "$1" "$target"/etc/group ; then
-      grep "$1:x:.*$USERNAME" "$target"/etc/group || $ROOTCMD adduser "$USERNAME" "$1"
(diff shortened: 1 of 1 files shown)


config/scripts/GRMLBASE/21-usersetup
```diff
@@ -31,7 +31,7 @@ fi
 add_user_to_group() {
    [ -n "$1" ] || return 1
    if grep -q "$1" "$target"/etc/group ; then
-      grep "$1:x:.*$USERNAME" "$target"/etc/group || $ROOTCMD adduser "$USERNAME" "$1"
+      $ROOTCMD adduser -q "$USERNAME" "$1"
```
mika, line 34: Uff, the adduser(8) manpage doesn't even mention `-q` and only documents:

```
       --quiet
              Synonymous to --stdoutmsglevel=warn. Deprecated.
```

:(

mika (approved):

config/scripts/GRMLBASE/21-usersetup
zeha: oh meh. well we can just use `--quiet`.

mika (approved):

PR #261, opened 2024-12-28
etc/fstab: drop examples handled by systemd

zeha:
These example entries are mounted by systemd. Avoid redefining them over systemds internal definition.

Mirrors https://github.com/grml/grml-udev-config/pull/4

Alternatively we could just drop the file from grml-live completely? grml-udev-config will write it anyway (I think).

### DIFF
--- a/config/files/etc/fstab/GRMLBASE
+++ b/config/files/etc/fstab/GRMLBASE
@@ -5,10 +5,6 @@
 # ${GRML_FAI_CONFIG}/files/etc/fstab/GRMLBASE
 #
 # <filesystem> <mountpoint>   <type> <options>                             <dump> <pass>
-proc           /proc            proc   rw,nosuid,nodev,noexec                0      0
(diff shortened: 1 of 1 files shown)


mika (approved):
Regarding:

> Alternatively we could just drop the file from grml-live completely? grml-udev-config will write it anyway (I think).

Yeah, that might be worth a look, though I *think* it's easier to customize /etc/fstab via grml-live than via the grml-udev-config Debian package (at least for *now* as it is :p)

zeha:
> Regarding:
> 
> > Alternatively we could just drop the file from grml-live completely? grml-udev-config will write it anyway (I think).
> 
> Yeah, that might be worth a look, though I _think_ it's easier to customize /etc/fstab via grml-live than via the grml-udev-config Debian package (at least for _now_ as it is :p)

Yes, but I think the customization doesn't live very long :-)

PR #263, opened 2024-12-30
grub: remove nonfunctional memtest86 bios-mode selection

zeha:
Because we do not have any cpu selection logic in the isolinux boot path, grml-live *always* installs the BIOS-mode memtest86+ binaries as /boot/addons/memtest.

Thus, the check for /boot/addons/memtest86+x64.bin and /boot/addons/memtest86+ia32.bin can never have worked. Drop it.

On `grml-small-daily20241223build179testing-amd64.iso`, only `Memory test (memtest86+)` is visible in BIOS-grub.

### DIFF
--- a/templates/boot/grub/addons.cfg
+++ b/templates/boot/grub/addons.cfg
@@ -25,31 +25,11 @@ fi
 
 # BIOS/non-EFI:
 if [ "${grub_platform}" != "efi" ] ; then
-  # try to detect amd64 by checking whether CPU supports 64-bit (long) mode
-  if cpuid -l ; then
-    if test -e /boot/addons/memtest86+x64.bin ; then
-      menuentry "Memory test (memtest86+x64.bin)" {
-        insmod linux16
-        linux16 /boot/addons/memtest86+x64.bin
-      }
-    elif test -e /boot/addons/memtest ; then # fallback to old memtest
-      menuentry "Memory test (memtest86+)" {
-        insmod linux16
-        linux16 /boot/addons/memtest
-      }
-    fi
-  else  # assume i386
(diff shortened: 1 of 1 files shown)


templates/boot/grub/addons.cfg
```diff
@@ -25,31 +25,11 @@ fi
```
mika, line 25: Hm, we have the same checks for `/boot/addons/memtest86+x64.efi` and `/boot/addons/memtest86+ia32.efi` also here for the EFI case, shouldn't this also be affected then?  :thinking: 

I agree with your conclusion in the commit message, but am pretty sure that this used to work for me in the past when I was working on this in e.g. https://github.com/grml/grml-live/commit/1a9a788fa46053e882e7332c6403b59ff8d437f1  + https://github.com/grml/grml-live/commit/c01a86b3fc9055a06f38db11b9e908b5fe03701f. Sadly I can't recall under which circumstances exactly I tested this, maybe also grml2usb was involved somehow? :thinking:  :disappointed:

zeha: No, grml-live does not rename the `.efi` files.

zeha: c01a86b sets it up as described:

```
        # provide memtest86+ >=6.00-1 files as "memtest" file
        # for BIOS boot in isolinux/syslinux
```

so the `*.bin` files always end up renamed to "memtest". I don't see how this can have worked in c01a86b.

mika: Ahhhh, I might have seen only the `Memory test (memtest86+x64.efi)` and didn't pay attention to efi vs bin naming, just like I missed also now. :smiling_face_with_tear: 
Thanks for your attention to detail!

PR #265, opened 2025-01-02
Workaround /sbin/init vanishing, Debian bug #1056151

zeha:
FAI before 6.2.3 diverts /sbin/init. Previously we got away with this by having systemd-sysv installed into the basefiles and then never upgrading it. But since c84d711d215eb62c1e4712f5280630000804a544 systemd-sysv gets installed in instsoft, triggering the broken behaviour.

Fixes: c84d711d215eb62c1e4712f5280630000804a544


zeha:
Please merge it if you deem it fine.

mika (approved):

PR #268, opened 2025-01-02
Install zsh completion into /usr/share instead of /etc

zeha:
Removes yet another dpkg conffile.


mika (approved):
Uh, shame on me that I didn't take care of this sooner myself :grinning: 
Thx!

zeha:
Meh, time is finite. Often there are more important things in life :)

PR #272, opened 2025-01-05
Use dropin config files for systemd, modules-load.d

zeha:
Avoid overwriting these files.

Also TIL: /etc/modules is deprecated!


zeha:
BTW, fuse is builtin. But I guess we don't want to rely on that not changing.

config/files/etc/systemd/logind.conf/GRMLBASE
```diff
@@ -1,23 +0,0 @@
...
-NAutoVTs=12
-#ReserveVT=6
-#KillUserProcesses=no
-#KillOnlyUsers=
-#KillExcludeUsers=root
-#InhibitDelayMaxSec=5
-#HandlePowerKey=poweroff
-#HandleSuspendKey=suspend
-#HandleHibernateKey=hibernate
-HandleLidSwitch=ignore
```
mika, line 15: I think the `HandleLidSwitch=ignore` is also relevant? Or put different: I think we don't want to suspend (notebook) systems by default in that situation, nor?

zeha: oops, i thought that line was commented out!

zeha: fixed now. thanks for the careful review!

mika (approved):
:rocket:

PR #278, opened 2025-01-06
Remove workaround for lighty issue from 15years ago

zeha:
http://trac.lighttpd.net/trac/ticket/657 was apparently fixed 15 years ago.


mika (approved):
:older_adult: :smile:

PR #281, opened 2025-01-06
Replace FAI with our own implementation

zeha:
Replace FAI with our own implementation

Just short of 700 new lines of Python, but covers running
mmdebstrap and lots of workarounds go away. I've titled it
"minifai".

`grml-policyrcd` gets removed, as minifai handles installing its own policy-rc.d.
This is necessary regardless if we want grml-policyrcd or not, and then grml-policyrcd provides no value anymore.

The variables `FAI_ARGS`, `FAI_DEBOOTSTRAP_OPTS` and `ZERO_FAI_LOGFILE` are no longer supported.
The grml-live parameter `-V` is now always enabled and the option is deprecated.

minifai logs to stdout/stderr, and aborts on the first error it encounters. All log messages now end up in grml-live.log.

These log/data files are no longer written:
- debconf.data,
- fai-monitor.log
- fai.conf
- fcopy.log
- nfsroot.conf
- scripts.log
- software.log
- status.log
- task_error
- task_local_error
- variables.log

Notably missing: packages can only be "install"-ed, which works like "install-norec" in FAI.  "hold", etc are not implemented at all.

Followup work: move script execution into the chroot where possible.


zeha:
@mika want to take a look? I need to work on the TODOs, but generally that's how it could look for now.

zeha:
cb pointed out minifai doesn't get installed into the deb. @zeha need to fix this

zeha:
> cb pointed out minifai doesn't get installed into the deb. @zeha need to fix this

done!

mika (approved):
This is fantastic on so many levels, many thanks and :bow: for this! :heart: 

We went through all the open questions and minor issues together onsite and I'd very much want to see this merged into master! :rocket:

PR #283, opened 2025-01-06
Fix missing isolinux files

zeha:
Fixes: 81efbabe7211084b9b75803c7040b929c15fbcce dbfb6371f64363805a1a96b628d626e34372839b

All `*.c32` files went missing in the ISO because the path calculation is wrong.


zeha:
Merged now so todays ISOs will hopefully work.

PR #285, opened 2025-01-08
Remove chainloading of grub from isolinux

zeha:
This drops support for booting grub on PC BIOS platforms. Grml ISOs boot using isolinux on this platform, but provided a chainloading option to use grub. Given all addons can now be booted from isolinux, I think there is no usecase left.


mika (approved):
Yes, let's drop this finally! There shouldn't be a use-case for legacy GRUB in 2025 anymore :partying_face:

PR #288, opened 2025-01-18
ZFS: derive linux package names from primary Debian architecture

zeha:
Necessary for building for arm64.


akorn:
I'd say this is a no-brainer in principle; go ahead.

(My only question is whether `dpkg --print-architecture` will always print what you mean; maybe there are some corner cases where you'd want to obtain the architecture some other way, or some architectures where the kernel packages are named differently? But that's a bridge to burn when we get there.)

zeha:
> or some architectures where the kernel packages are named differently?

these might exist, yes

> But that's a bridge to burn when we get there.

indeed, i'd worry about it when we support such an arch.

PR #294, opened 2025-01-30
Fix timezone and RTC setup

zeha:
Correctly create /etc/localtime as a symlink. Also stop dealing with /etc/adjtime, which does not exist anymore. Instead document RTC=UTC as default.

Remove GRMLBASE.var "UTC", and default TIMEZONE to Etc/UTC, to get the debconf variables correctly set.

Corresponding grml-autoconfig change: https://github.com/grml/grml-autoconfig/pull/30

### DIFF
--- a/config/class/GRMLBASE.var
+++ b/config/class/GRMLBASE.var
@@ -3,6 +3,4 @@
 # allow installation of packages from unsigned repositories
 # FAI_ALLOW_UNSIGNED=1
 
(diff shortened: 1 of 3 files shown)


jkirk:
FTR, related: https://github.com/grml/grml-autoconfig/pull/30

jkirk (approved):

mika (approved):

Issue #296, opened 2025-01-30
Review help&info texts in grub boot menu


jkirk:
Triggered by the boot options changes via grml/grml-live#294 + grml/grml-autoconfig#30, the following options must be removed:

- [x] `grml gmt tz=Europe/Vienna` in 
 
https://github.com/grml/grml-live/blob/f5b89ad898a7c107de9a5d9d19fc4dc99d930e75/templates/boot/isolinux/f4#L12

We also need to review the other boot options listed in the isolinux bootsplash screens, related to grml/grml-live#243.

jkirk:
The following options in the isolinux bootsplash screens f4 + f5 need to be reviewed:

- [ ] grml blacklist=firewire-core               disable new (Juju) firewire stack
- [ ] grml blacklist=ieee1394                           disable old firewire stack
- [ ] grml blacklist=ipw3945,prism54                     disable specified drivers
- [ ] grml blacklist=module1[,module2]  disable loading of specified module, e.g.:
- [ ] grml blacklist=usbcore                                   disable USB drivers
- [ ] grml debnet        search local partitions for file /etc/network/interfaces,
- [ ] grml localtime                   configure Hardware Clock / RTC to localtime
- [ ] grml tz=Europe/Vienna                         set timezone (default: tz=UTC)
- [ ] grml home=/dev/sda1                   use specified device as home directory
- [ ] grml keyboard=de                                                set keyboard
- [ ] grml lang=[at|ch|da|de|es|fr|it|...|us]              set keyboard + language
- [ ] grml myconfig=/dev/fd0              load configuration from specified device
- [ ] grml noautoconfig               disable searching for device labeled GRMLCFG
- [ ] grml nodhcp                     disable searching for network setup via DHCP
- [ ] grml noeject noprompt      do not eject CD after halt / do not prompt for it
- [ ] grml nofstab / nolabel   disable generating of /etc/fstab / use of fs-labels
- [ ] grml nolvm                          disable search for Logical Volumes (LVM)
- [ ] grml noquick                         disable grml-quickconfig startup script
- [ ] grml noswraid                       disable search for software RAID devices
- [ ] grml noudev                disable startup of udev (disables module loading)
- [ ] grml services=foo1[,foo2] start specified service(s) [/etc/init.d/name foo1]
- [ ] grml ssh=password     set password for user grml + root and start ssh-server
- [ ] grml startx[=windowmanager]                     autostart X.org using grml-x
[...]

PR #298, opened 2025-01-30
isolinux: Drop bootoption "gmt"

jkirk:
We replaced "gmt" with "utc" some time ago, see: 65ee43c.

In the meanwhile we changed the timezone and localtime handling and dropped support for the bootoption "utc", as it is the default nowadays, see: grml/grml-autoconfig@e96da70 + grml/grml-live@34167e2f.

TZ defaults to UTC and the RTC defaults to be set to UTC.

Related: grml/grml-live#296

### DIFF
--- a/templates/boot/isolinux/f4
+++ b/templates/boot/isolinux/f4
@@ -9,7 +9,8 @@
  grml blacklist=usbcore                                   disable USB drivers  
  grml debnet        search local partitions for file /etc/network/interfaces,  
                 copy /etc/network then to local system and restart networking  
- grml gmt tz=Europe/Vienna   use GMT-based time and specified timezone for TZ  
+ grml localtime                   assume Hardware Clock (RTC) is in localtime  
+ grml tz=Europe/Vienna                         set timezone (default: tz=UTC)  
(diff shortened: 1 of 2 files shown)


templates/boot/isolinux/f4
```diff
@@ -9,7 +9,8 @@
  grml blacklist=usbcore                                   disable USB drivers  
  grml debnet        search local partitions for file /etc/network/interfaces,  
                 copy /etc/network then to local system and restart networking  
- grml gmt tz=Europe/Vienna   use GMT-based time and specified timezone for TZ  
+ grml localtime                   configure Hardware Clock / RTC to localtime
```
zeha, line 12:
```suggestion
 grml localtime                   assume Hardware Clock (RTC) is in localtime  
```

mika (commented):
LGTM with the suggestion from Chris!

templates/boot/isolinux/f4
jkirk: Thx for the suggestion. Adjusted the wording in grml-cheatcodes.txt + updated the PR. Please re-review.

mika (approved):

PR #300, opened 2025-01-30
Merge grml-autoconfig

zeha:
I ended up doing a manual `git mv` and fixed up the remaining issues in the merge commit. Seemed more true than using git-filter-repo.


zeha:
https://github.com/grml/grml/issues/225#issue-2841854235

zeha:
Needs https://github.com/grml/grml-scripts/pull/23

zeha:
grml-autoconfig had shunit tests, but:
* they were disabled since 2019
* are written in zsh, which shellcheck cannot handle

I'd say we merge them, and either fix or delete them later.

Issue #267, opened 2025-01-02
RFC: Redesign of grml-live


mika:
# Redesign of grml-live

## Main tasks of grml-live

grml-live is a build system for creating a Grml and Debian based Linux Live system.

grml-live includes the following main stages:

* FAI dirinstall: [FAI](http://fai-project.org) (Fully Automatic Installation) does the chroot related work with hooks/scripts provided by grml-live (related to `grml_chroot` directory)
* mksquashfs: generates the compressed squashfs file which provides the chroot in compressed format (related to `grml_cd` directory)
* xorriso: generates the ISO itself, based on the output by mksquashfs and some further bootmanger related stuff handled by grml-live (related to `grml_iso` directory)

We don't plan to replace any of `mksquashfs` nor `xorriso`, though consider getting rid of `FAI` (with its `dirinstall` feature and its `fcopy` tool that we currently use within grml-live).

## Rational for getting rid of FAI

FAI is a tool for unattended mass deployment of Linux, which is available since 1999. Especially the class system of FAI was a compelling and promising reason to build grml-live on top of FAI. grml-live therefore was based on FAI since its early beginnings in 2007.

grml-live was designed for building Debian based live systems and relied on FAI's `dirinstall` feature underneath. The way live systems are built might look similar to disk based installations. But with grml-live we have specific needs for our release process that FAI was never designed for. We also provide ISO specific customization options and need to comply with GPL needs. Neither of those requirements are fullfilled by FAI.

We also had to deal with behavior changes within FAI (e.g. shell.log vs. scripts.log: https://github.com/grml/grml-live/commit/74c4bce9cc66d70eff3023b8a048d0228869a12e, make-fai-nfsroot.conf: https://github.com/grml/grml-live/commit/75e65f36279dcb8941137bc45c806392b27e3649, ignoring packages: https://github.com/grml/grml-live/commit/121b3484e6301de73a1c5f3cc89ba4997de9ba08).

As with any software, we hit bugs also with FAI, like e.g. [#760133](https://bugs.debian.org/760133) (`'mount --make-runbindable' breaks underlying device for unrelated operations`), [#989547](https://bugs.debian.org/989547) (`ROOTCMD relies on specific unshare features`), and [#1056151](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1056151) (`which makes /usr/sbin/init vanish in /usr-move conditions`).

[...]

PR #301, opened 2025-01-30
Provide migration script migrate-to-minifai

mika:
Executing it converts our existing config/* from FAI layout (e.g. config/files/$FILENAME/$CLASS) into the new minifai layout (e.g. config/files/$CLASS/$FILENAME).

See https://github.com/grml/grml-live/issues/267

PR #302, opened 2025-01-30
Adopt new layout for files/, env/, hooks/

zeha:
Implements @akorn's "alternate layout" from #267


config/files/GRMLBASE/usr/share/initramfs-tools/scripts/init-top/grml
akorn, line 1: Those carriage returns should probably be restored. :)

akorn: Do you think it's worthwhile to avoid the literal `^M` by introducing `CR="$(printf '\015')"`, and then just using `$CR` instead of `^M` everywhere? This should avoid confusing editors.

usr/lib/grml-live/minifai
akorn, line 1: Maybe it's just me, but I find the fcopy logic hard to follow without comments. The function and variable names aren't really intuitive for me.

mika (commented):
You used my migration script from https://github.com/grml/grml-live/pull/301, right?

Basically LGTM, if we're all fine with switching to this approach.
We should probably provide proper documentation and debian/NEWS information regarding this change?

mika:
> I used the script from #301 yes. Maybe you want to put NEWS and migration advice into #301?

ACK, good idea!

> Regarding docs, we'll need general docs on how things work once we've walked away from how FAI works. Not sure where to put it; probably a new file in docs/?

Indeed, that sounds like a good idea as well!

PR #303, opened 2025-01-30
SW: add gnupg to GRML_FULL

mika:
Now that we dropped gnupg as dependencty from the grml-keyring package, we need to make sure to have it available at least on GRML_FULL.

We might consider migrating to sq instead in the future, feedback welcome.


zeha (approved):

mika:
FTR, build failure only due to underlying mirror changes, not related to actual change

PR #307, opened 2025-02-04
Follow ipxe 1.21.1+git20220113.fbbdc3926+dfsg-3 changes

zeha:
Only supports the new filename scheme for EFI; for legacy boot nothing has changed.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -1045,8 +1045,14 @@ else
     einfo "Installing boot addons."
 
     # copy from chroot-installed packages
-    copy_file_logged "${BUILD_OUTPUT}"/boot/addons/ipxe.lkrn "${CHROOT_OUTPUT}" /usr/lib/ipxe/ipxe.lkrn
-    copy_file_logged "${BUILD_OUTPUT}"/boot/addons/ipxe.efi "${CHROOT_OUTPUT}" /usr/lib/ipxe/ipxe.efi
+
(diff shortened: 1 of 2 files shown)


grml-live
```diff
@@ -1045,8 +1045,14 @@ else
     einfo "Installing boot addons."
 
     # copy from chroot-installed packages
-    copy_file_logged "${BUILD_OUTPUT}"/boot/addons/ipxe.lkrn "${CHROOT_OUTPUT}" /usr/lib/ipxe/ipxe.lkrn
-    copy_file_logged "${BUILD_OUTPUT}"/boot/addons/ipxe.efi "${CHROOT_OUTPUT}" /usr/lib/ipxe/ipxe.efi
+
+    # ipxe >= 1.21.1+git20220113.fbbdc3926+dfsg-3
```
mika, line 1049: hm, you think it's not worth supporting bookworm based releases?

zeha: I think it's a bit of a niche feature. We can support old releases if someone (you?) cares?

zeha:
NB: some of the "files" in the new ipxe packaging are symlinks to files in /boot. This is okay, if surprising.

grml-live
mika: Fair enough, ack :)

mika (approved):

PR #308, opened 2025-02-08
Document kernel option dis_ucode_ldr

mika:
Booting a live system probably shouldn't trigger any microcode updates. The dis_ucode_ldr kernel parameter disables the microcode loader, see https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html and
https://wiki.debian.org/Microcode#Working_around_boot_problems_caused_by_microcode_updates

### DIFF
--- a/templates/GRML/grml-cheatcodes.txt
+++ b/templates/GRML/grml-cheatcodes.txt
@@ -230,6 +230,7 @@ grml libata.force=[ID:]VAL            Force configurations for libata.
                                       to force pio4 mode on device "ata1:00:"
 grml libata.dma=0                     Disable DMA on PATA and SATA devices
 grml libata.ignore_hpa=1              Disable host protected area (which should enable the whole disk)
+grml dis_ucode_ldr                    Disable loading of CPU microcode
 grml vga=normal                       No-framebuffer mode (does not influence X)
 grml vga=ask                          Display menu for framebuffer mode
 grml radeon.modeset=0  nomodeset      Disable Kernel Mode Setting (KMS) for Radeon driver.


evgeni:
Why not?

The microcode loading by the kernel is ephemeral, so you're not altering the system.

You could argue you want a system that matches the original as closely as possible (for debugging etc), but then again you're already not getting that as either the microcode in the bios is new enough, then no loading is happening or the bios is old and the to-be-debugged OS brings updates that are not loaded by grml.

And then having as new microcode as possible seems like a good idea?

mika:
> Why not?
> 
> The microcode loading by the kernel is ephemeral, so you're not altering the system.
> 
> You could argue you want a system that matches the original as closely as possible (for debugging etc), but then again you're already not getting that as either the microcode in the bios is new enough, then no loading is happening or the bios is old and the to-be-debugged OS brings updates that are not loaded by grml.
> 
> And then having as new microcode as possible seems like a good idea?

Good point, thanks!
The issue came up for me when I noticed that Debian's `recovery mode` options in GRUB uses `dis_ucode_ldr` nowadays.

Adding yet another separate boot menu entry is also not optimal though, I guess?

Also, quoting from https://wiki.debian.org/Microcode#Working_around_boot_problems_caused_by_microcode_updates:

> Very rarely, it is possible for a kernel or a microcode update bug to cause boot issues (hangs or resets at the very beginning of the boot process) on specific processor models.

> For this reason, it is possible to pass a [kernel command line parameter](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html) that skips the microcode update process entirely. The parameter to disable microcode loading is dis_ucode_ldr.

I was also not aware of all the information that's available from within http://metadata.ftp-master.debian.org/changelogs/non-free/i/intel-microcode/unstable_README.Debian, though the `Triggering an immediate microcode update (without a reboot)` section feels like that's not recommended either. :-/

Has anyone of you observed problems with microcode updates in the wild and would have preferred the automatic updates on Grml to be disabled (by default)? :thinking:

evgeni:
Right, but Debians recovery mode needs to recover you from a bad Debian update, there it makes sense to not load a possibly faulty microcode (not that I've ever had to do that, but  that's a different question).

Grmls microcode will be (maybe) different from the OS'es as it's snapshotted at a different time.

I'd probably just prominently document that option and not worry much further

zeha:
We might also consider not shipping microcode at all. Depends a bit on the release cadence and if we expect people to use old ISOs.

zeha:
> We might also consider not shipping microcode at all. Depends a bit on the release cadence and if we expect people to use old ISOs.

So it turns out today we do not ship any microcode.

templates/GRML/grml-cheatcodes.txt
```diff
@@ -230,6 +230,7 @@ grml libata.force=[ID:]VAL            Force configurations for libata.
                                       to force pio4 mode on device "ata1:00:"
 grml libata.dma=0                     Disable DMA on PATA and SATA devices
 grml libata.ignore_hpa=1              Disable host protected area (which should enable the whole disk)
+grml dis_ucode_ldr                    Disable the microcode loader
```
zeha, line 233:
```suggestion
grml dis_ucode_ldr                    Disable loading of CPU microcode
```
?

Or

```suggestion
grml dis_ucode_ldr                    Disable CPU microcode loader
```

zeha (approved):

evgeni (approved):

templates/GRML/grml-cheatcodes.txt
mika: Good idea, thx - done

zeha (approved):

PR #311, opened 2025-03-06
SW: add iwd to GRMLBASE

mika:
iwd seems to be required for support of wpa-enterprise networks. Given that we consider network connectivity as essential and already ship the wpasupplicant package in GRMLBASE, and iwd only adds 2,717 kB of additional disk space (including its dependency libell0), let's also include it into our base software selection.

Closes: grml/grml#226

### DIFF
--- a/config/package_config/GRMLBASE
+++ b/config/package_config/GRMLBASE
@@ -31,6 +31,7 @@ hwinfo
 ifupdown
 initramfs-tools
 iputils-ping
+iwd
 kbd
 less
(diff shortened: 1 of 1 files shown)


mika:
I don't know whether wireless-regdb is needed? :thinking:

zeha:
> I don't know whether wireless-regdb is needed? :thinking: 

Probably?

zeha (approved):
ship it

jkirk (approved):

PR #314, opened 2025-03-13
usr/share/zsh/vendor-completions/_grml-live: use python instead of jq

crpb:
* jq needs to be installed, python3 most likely not. \o/
  * and don't forget about curl...

### DIFF
--- a/usr/share/zsh/vendor-completions/_grml-live
+++ b/usr/share/zsh/vendor-completions/_grml-live
@@ -58,11 +58,16 @@ _grmllive_waybackdates() { #{{{
   local expl
   local -a waybackdates 
 
-  if command -pv jq curl >/dev/null; then
+  getdates() {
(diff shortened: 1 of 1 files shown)


usr/share/zsh/vendor-completions/_grml-live
```diff
@@ -58,11 +58,16 @@ _grmllive_waybackdates() { #{{{
   local expl
   local -a waybackdates 
 
-  if command -pv jq curl >/dev/null; then
+  getdates() {
+    whence -p python3 || return 0
```
mika, line 62: I assume this one should be `whence -p python3 &>/dev/null || return 0`, otherwise python3 also shows up in the list? :)

mika (changes requested):
Good idea, thanks!

mika (approved):
LGTM (thanks @crpb!), leaving +2 to @zeha :)

Issue #312, opened 2025-03-10
fcopy fails when overwriting symlinks?


zeha:
https://github.com/grml/grml-live/actions/runs/13664433866/job/38202659919

```
I: fcopy: Installing /source/config/files/etc/apt/sources.list.d/grml.sources/GRMLBASE as /tmp/tmp1324qvwy/grml_chroot/etc/apt/sources.list.d/grml.sources.
E: 2025-03-04T22:52:41.883693 helper_socket_thread caught fatal exception
Traceback (most recent call last):
  File "/source/usr/lib/grml-live/minifai", line 391, in helper_socket_thread
    rc = do_fcopy(conf_dir, chroot_dir, classes, req[1:])
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/usr/lib/grml-live/minifai", line 341, in do_fcopy
    do_fcopy_path(files_dir / dir, chroot_dir / dir, classes, mode)
  File "/source/usr/lib/grml-live/minifai", line 264, in do_fcopy_path
    shutil.copyfile(to_copy, dest_path, follow_symlinks=False)
  File "/usr/lib/python3.11/shutil.py", line 258, in copyfile
    with open(dst, 'wb') as fdst:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp1324qvwy/grml_chroot/etc/apt/sources.list.d/grml.sources'
E: minifai fcopy: 2025-03-04T22:52:41 got no reply from server
E: Script /source/config/hooks/updatebase.GRMLBASE failed with exitcode 119 - aborting.
```

Needs investigation if this is a general problem with pre-existing symlinks.

zeha:
The situation might have been:

- grml-live in run 1 installs a file
- it then installs a package overwriting this file with a symlink
- in run 2, grml-live wants to overwrite the symlink again with a file
- this fails

mika:
FTR, now with the grml-keyring situation fixed, also the tests pass (see https://github.com/grml/grml-live/pull/311)

PR #316, opened 2025-03-31
fcopy: remove target files and symlinks before overwriting them

zeha:
Closes: grml/grml-live#312.


zeha:
Haven't tested this, but I think it's correct...

Issue #306, opened 2025-01-30
config/files/GRMLBASE/usr/share/initramfs-tools/scripts/init-top/grml hard to edit


zeha:
`config/files/GRMLBASE/usr/share/initramfs-tools/scripts/init-top/grml` has embedded control characters (especially ^M etc) which confuse vim, neovim, zed and git diff.
Would be good if we can avoid having them in the file.

mika:
I like @akorn's suggestion regarding using something like `CR="$(printf '\015')"` and have `$CR` everywhere `^M` is needed.

PR #317, opened 2025-03-31
Stop inlining \r and ESC in init-top/grml

zeha:
Closes: #306


mika (approved):
Nice, thanks! Not verified on my own, let's keep an eye on boot process to ensure every behaves as expected :)
Thx for taking care! :heart:

PR #318, opened 2025-04-13
CLASS SYSTEMD_NETWORKD

crpb:
* Add systemd-networkd class and a default dhcp-fallback configuration to catch any unconfigured network interfaces which might likely be the case.

### DIFF
--- a/config/files/SYSTEMD_NETWORKD/etc/systemd/network/99-dhcp-fallback.network
+++ b/config/files/SYSTEMD_NETWORKD/etc/systemd/network/99-dhcp-fallback.network
@@ -1,0 +1,13 @@
+# We will match any ethernet adapter that is left unconfigured.
+[Match]
+Type=ether
+
+[Network]
+DHCP=yes
+
+[DHCPv4]
+# In our DHCP we trust
+UseDomains=true
+# Don't use duid
+ClientIdentifier=mac
+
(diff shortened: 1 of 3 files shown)


mika (commented):
I like it, @zeha what's your take on this?

zeha:
> I like it, @zeha what's your take on this?

I think it conflicts with what we're doing in grml-network, etc.

I'm not sure I understand the usecase. If it stays off by default, it doesn't provide much value. If it becomes on by default, we should drop grml-network and so on.

zeha:
> Good point indeed, we should discuss whether and if at all, how we could move towards systemd-networkd  ...

Let's do a release, then merge autoconfig et al into grml-live, and maybe then all of this becomes easier.

crpb:
btw, i only came up with that as i was irritated by dhcpcd's "duid" default. i tried to implement a cheatcode but grml-autoconfig is running to late from what i can tell so far

https://github.com/grml/grml-live/compare/master...crpb:grml-live:dhcpcdclientid

anarcat:
> btw, i only came up with that as i was irritated by dhcpcd's "duid" default.

now i'm curious, what's that?

crpb:
> now i'm curious, what's that?

https://manpages.debian.org/testing/dhcpcd-base/dhcpcd.8.en.html#D
https://manpages.debian.org/testing/dhcpcd-base/dhcpcd.conf.5.en.html#clientid

Issue #73, opened 2019-03-08
SSH server boot: enable announce


lars18th:
Hi,

The current `ssh` cheat code doesn't activate any option to discover the server in the network (it only prints in the console the message `You can connect with SSH to: ...`. And that's useless with a headless machine.

However, the GRML Live has the AVAHI-DAEMON service!
So with these commands you can announce the server in the network:
```
cp /usr/share/doc/avahi-daemon/examples/ssh.service /etc/avahi/services
systemctl enable avahi-daemon.service
systemctl start avahi-daemon.service
```

So, I suggest to do this _every time_ the `ssh` cheat code is used. It's free and easy to activate!

Futhermore, you can enhance it publising the password (as an option) in the Name of the server (now the default is `grml`; so you can use `grml (pw:<password>)`

I hope you aggree with this enhancement. It will simplifies the connection of a headless server booted with GRML Live.

jkirk:
Hi!

Thank you for your suggestion. I like the idea. I haven't used the avahi-utils for quite some time now. But do I "guess"  right (I just looked up an example :)), that one could then discover the SSH Service with the following command: `avahi-browse -d local _ssh._tcp --resolve -t`?

Apart from that I am not quite sure if publishing the password is good idea. Why would you want that? The password is set on boot, so it should be known to the user - and should be changed as soon as possible, of course.

/cc @mika What do you think?

mika:
I like the suggestion WRT avahi, we should definitely give this a try!

Announcing the password calls for security troubles though, that's nothing I would consider to enable as default.

mika:
This is wonderful idea @lars18th, I really like it, thank you very much for raising this idea here! 🙇 
@jkirk implemented this for our upcoming Grml release 2025.05, thanks! 👍

mika:
> Now another question to address: it's about the ssh password. Why not use something easy to remember/generate? For example: `grml-<MAC_ADDR>` or something similar? What you think?

I like your idea! :) But *which* MAC-Address if there are multiple network devices? 🤔 

We definitely also should make the password more visible, esp. as it's easy to overlook on the boot splash or even press a key on the tty1 and then the password isn't visible at all anymore. We could f.e. integrate it also within `grml-quickconfig`?

What might also make life easier if we use something like `xkcdpass` for generating the random password:

```
% xkcdpass --count=1 --acrostic='grml' --min=2 --max=6              
grass reopen mop lushl

% xkcdpass --count=1 -n 2 --delimiter=G
[...]

PR #327, opened 2025-05-14
Announce SSH service with Avahi

jkirk:
We want to announce the SSH service with Avahi.

The avahi-daemon is shipped in GRML_FULL but needs be enabled and a Avahi ssh service configuration file needs to be deployed.

The avahi-ssh.service deploys the SSH service config for Avahi when the SSH service is started. The service avahi-ssh.service needs to be enabled too.

To discover the SSH service in the local network the following command can be used:

```
  avahi-browse -d local _ssh._tcp --resolve -t
```

Issue: grml/grml-live#73

### DIFF
--- a/config/files/GRML_FULL/etc/systemd/system/avahi-daemon.service.d/override.conf
+++ b/config/files/GRML_FULL/etc/systemd/system/avahi-daemon.service.d/override.conf
@@ -1,0 +1,4 @@
+# This file was deployed via grml-live.
+
+[Install]
+WantedBy=grml-boot.target
--- a/config/files/GRMLBASE/etc/systemd/system-preset/10-grml.preset
+++ b/config/files/GRMLBASE/etc/systemd/system-preset/10-grml.preset
@@ -8,5 +8,7 @@ enable gpm.service
 enable grml-autoconfig.service
 enable debug-shell.service
 enable resolvconf.service
+enable avahi-daemon.service
+enable avahi-ssh.service
 
 disable *
--- a/config/files/GRMLBASE/etc/systemd/system/avahi-ssh.service
+++ b/config/files/GRMLBASE/etc/systemd/system/avahi-ssh.service
@@ -1,0 +1,15 @@
+# This file was deployed via grml-live.
+
+[Unit]
+Description=Deploy SSH service config for Avahi
+ConditionPathExists=/usr/share/doc/avahi-daemon/examples/ssh.service
+ConditionPathExists=/etc/avahi/services
+PartOf=ssh.service
+After=ssh.service
+
+[Service]
+Type=oneshot
+ExecStart=cp /usr/share/doc/avahi-daemon/examples/ssh.service /etc/avahi/services/
+
+[Install]
+WantedBy=ssh.service


zeha:
./config/files/GRMLBASE/etc/systemd/system-preset/10-grml.preset

zeha:
please add `# This file is installed by grml-live.` as a header

zeha (approved):

config/files/GRML_FULL/etc/systemd/system/avahi-ssh.service
```diff
@@ -0,0 +1,15 @@
+# This file is installed by grml-live.
```
mika, line 1: nit: we tend to use `# This file was deployed via grml-live.`, maybe keep it consistent?

mika, line 1: And also, given that we enable `avahi*` stuff in GRMLBASE, maybe also move that file to `GRMLBASE` so it's not a grml-full specific feature?

config/files/GRML_FULL/etc/systemd/system/avahi-ssh.service
```diff
@@ -0,0 +1,15 @@
...
+[Unit]
+Description=Deploy SSH service config for Avahi
+ConditionPathExists=/usr/share/doc/avahi-daemon/examples/ssh.service
+ConditionPathExists=/etc/avahi/services
+PartOf=ssh.service
+After=ssh.service
+
+[Service]
+Type=oneshot
+ExecStart=cp /usr/share/doc/avahi-daemon/examples/ssh.service /etc/avahi/services
```
mika, line 12: Maybe use `/etc/avahi/service/` with trailing slash to clarify it's meant as directory (and not ends up as file `/etc/avahi/services`, even though the `ConditionPathExists` exists (while it might still refer to a file :p)?

mika (commented):
nitpick for commit message (I stil don't see any way how to comment git commit messages in github :-/): s/and a Avahi ssh service/and an Avahi ssh service/

I like it, only nitpicks, thanks! :+1:

config/files/GRML_FULL/etc/systemd/system/avahi-ssh.service
zeha: ah, but i copied this text from another file from grml-live :D

mika: Yeah sorry, inconsistency ftw 8-)

```
% git grep '# This file was deployed via grml-live' | wc -l
28
% git grep '# This file is installed by grml-live' | wc -l
1
```

jkirk: Hmm. `avahi-daemon` is deployed in GRML_FULL (only):
```
❯ git grep avahi-daemon | grep package_config
config/package_config/GRML_FULL:avahi-daemon
```

Does it make sense to install the `avahi-ssh.service` in GRMLBASE then?

zeha:
```
$ git grep 'This file is installed by grml-live' | wc -l
       2
```

😆

config/files/GRML_FULL/etc/systemd/system/avahi-daemon.service.d/override.conf
```diff
@@ -0,0 +1,4 @@
+# This file was deployed via grml-live.
```
mika, line 1: FTR: this file could *possibly* also go to `GRMLBASE` instead of `GRML_FULL`, but let's be on the safe side for the current release and let's look into that once grml-live incoporates also grml-autoconfig etc to see how this could be useful for every user of grml-live (who might not use `GRML_FULL` as base but e.g. `GRML_SMALL` or even just `GRMLBASE`) without breaking anything on non-GRML_FULL :)

zeha:
Screenshot 2025-05-14 at 17 20 00

works for me

zeha (approved):

PR #329, opened 2025-05-16
debian/grml-live.maintscript: drop leftover deprecated files

mika:
Those configuration files weren't properly removed with the removal of lenny, squeeze, wheezy, jessie, stretch + buster Debian release support within commit 5a77db439a8a43d9f07e5f4ec5e4ac500fea8be8


debian/grml-live.maintscript
```diff
@@ -3,12 +3,18 @@ rm_conffile /etc/grml/fai/config/debconf/GRMLBASE 0.51.0~
 rm_conffile /etc/grml/fai/config/files/etc/apt/apt.conf.d/15grml-live/GRMLBASE 0.51.0~
 rm_conffile /etc/grml/fai/config/files/etc/apt/apt.conf.d/20pdiffs/GRMLBASE 0.51.0~
 rm_conffile /etc/grml/fai/config/files/etc/apt/grml.key/GRMLBASE 0.32.3~
+rm_conffile /etc/grml/fai/config/files/etc/apt/preferences.d/jessie-backports/DEBIAN_JESSIE 0.43.0~
```
zeha, line 6: this will need to be 0.53.2~, no?

mika: uh right, I though because it got removed with version v0.43.0 that this would be the proper thing to do, but you're absolutely right, thanks! :bow:  Done in new upload!

zeha (approved):

PR #334, opened 2025-05-18
minifai: fallback to debian-archive-keyring

crpb:
* if GRML_FAI_CONFIG isn't in the same tree-depth as the original the relative links in buildstrap-keyring (blame debhelper!) will not be resolveable which makes mmdebstrap fail with keyring=None.

### DIFF
--- a/usr/lib/grml-live/minifai
+++ b/usr/lib/grml-live/minifai
@@ -608,6 +608,8 @@ def install_base(conf_dir: Path, chroot_dir: Path, classes, debian_suite: str, m
     for class_name in classes:
         if (keyring_dir / class_name).exists():
             keyring_file = keyring_dir / class_name
+        else:
+            keyring_file = '/usr/share/keyrings/debian-archive-keyring.pgp'
 
     args = [
         "mmdebstrap",


zeha:
imo the right thing to do is to fail early in this case.

zeha:
If someone provides a short reproducer, I can try to take a closer look.

mika:
Friendly ping /cc @youam - would appreciate your feedback here also!

youam:
Hey there. Thanks @mika for the nudge; otherwise this would've stayed on my todo-list for a while..

Reproducing this bug is really simple. Boot a fresh grml-full Nudlaug, prepare with the following steps:
```
cd /mnt
rsync -a /usr/share/grml-live ./
echo GRML_FAI_CONFIG=/mnt/grml-live/config > /etc/grml/grml-live.local
```
and grml-live will fail in the mmdebstrap step

crpb:
yeah, like that or via overlay so you don't have to update/merge the whole tree every time.
`/etc/fstab`
```
overlay /grml/config overlay x-systemd.automount,x-systemd.mount-timeout=10,x-systemd.idle-timeout=30,lowerdir=/usr/share/grml-live/config,upperdir=/grml/overlay/local,workdir=/grml/overlay/work 0 2
```

mika:
I ran into a similar situation, and as it turned out (and thanks to a hint by @zeha), part of my working directory was not accessible for others (as in: directory permissions of `750`), even if grml-live is executed as root.

@crpb might this also be related to the situation you experienced? :thinking:

PR #337, opened 2025-06-01
Redefine minifai actions

zeha:
Instead of the traditional dirinstall and softupdate FAI_ACTIONs, minifai now provides four:

* bootstrap: create chroot and install packages ("updatebase", "instsoft")
* dirinstall: everything: create chroot, install packages and run configure scripts
* softupdate: everything except creating the chroot
* reconfigure: only run configure scripts

Instead of distributing the logic of what to run when in grml-live, FAI and instsoft/updatebase hooks, grml-live now makes a decision and calls minifai with the correct action.

Notably, except for "reconfigure", the "updatebase" task now gets to always run, ensuring APT package lists are available. minifai's updatebase was already minimal and only runs `apt-get update`, which is what grml-live needs.

### DIFF
--- a/config/hooks/GRMLBASE/instsoft
+++ b/config/hooks/GRMLBASE/instsoft
@@ -15,15 +15,6 @@ set -e
 # FAI sets ${target}, but shellcheck does not know that.
 target=${target:?}
 
-# if hooks/updatebase.GRMLBASE fails for whatever reason
-# and can't skip instsoft.GRMLBASE we have to make sure
-# we exit here as well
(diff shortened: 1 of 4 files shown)


zeha:
The number of changed lines is not high, but it might have big consequences. @mika please think this through :-)

mika (approved):
Thanks, I'm all for making our life easier with the logic, esp. with softupdate and updatebase/instsof in `config/hooks/GRMLBASE/instsoft`.

mika (approved):

Issue #333, opened 2025-05-17
Automatically add `ca-certificates` if sources use https://


zeha:
For debian mirrors configured in `FAI_DEBOOTSTRAP` https:// works, but when extra sources are added using the files/.../etc/apt/sources.list.d/ mechanism, ca-certificates is not added. And there is no hook which could make this possible.

Reported by josch.

zeha:
I was going to add this code to `hooks/GRMLBASE/updatebase`, but I doubt it will work:

```bash
if grep -r https:// /etc/apt/sources.list* ; then
  echo "Ensuring ca-certificates is available for https:// mirrors"
  apt satisfy -q -y --no-install-recommends ca-certificates
fi
```

Problem that I see: when the updatebase hook runs for softupdate, there will be no apt cache. Thus, ca-certificates cannot be installed.
Another option might be to run `apt-get update` before `fcopy -M -i -B -v -r /etc/apt`, but it would need to be guarded as `if !softupdate`, and then we run `apt-get update` twice.

Maybe it'd be better to implement this inside minifai.

PR #338, opened 2025-06-01
minifai: always install ca-certificates to allow https:// sources

zeha:
Closes: grml/grml-live#333

### DIFF
--- a/usr/lib/grml-live/minifai
+++ b/usr/lib/grml-live/minifai
@@ -598,9 +598,9 @@ def install_base(conf_dir: Path, chroot_dir: Path, classes, debian_suite: str, m
     # Work around APT bug: http://bugs.debian.org/1092164
     included_packages = ["netbase"]
 
-    # Allow using https:// mirror, if given.
-    if mirror_url.startswith("https"):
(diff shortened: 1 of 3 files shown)


usr/lib/grml-live/minifai
```diff
@@ -598,9 +598,9 @@ def install_base(conf_dir: Path, chroot_dir: Path, classes, debian_suite: str, m
     # Work around APT bug: http://bugs.debian.org/1092164
     included_packages = ["netbase"]
 
-    # Allow using https:// mirror, if given.
-    if mirror_url.startswith("https"):
-        included_packages.append("ca-certificates")
+    # Allow using https:// sources. Do this unconditionally, so sources added with
+    # fcopy /etc/apt just work.
+    included_packages.append("ca-certificates")
```
evgeni, line 603: I think placing that into package_config/GRMLBASE would avoid relying in implementation details in minifai :)

zeha: No, package_config/GRMLBASE is applied too late: its contents are only installed after the initial chroot creation, and after task updatebase (calling `apt-get update`).

Issue #91, opened 2020-06-26
Support snapshot usage with -w <date> option also during bootstrapping


mika:
The wayback option `-w` doesn't yet respect the given date for bootstrapping the issue.
This can be worked around via providing a custom BASEFILE, though we should support the wayback option also during the automated debootstrap/mmdebstrap riun if no BASEFILE exists.

(Forwarding from customer, TT#69519 in customer's ticket system)

mika:
I think this now works in the initial bootstraping thanks to `mmdebstrap`!

Executing `grml-live -s stable -a amd64 -c GRMLBASE,GRML_SMALL,AMD64 -o [...] -v 2025-06-06 -w 20250601` v0.53.1 works fine in bootstrapping, now only failing later on:

```
[...]
Wayback date '20250601' identified, enabling for snapshot.debian.org usage.
D: fcopy recursive=True ignore_missing=True user='root' group='root' mode=420 paths=['etc/apt']
I: fcopy: Installing /usr/share/grml-live/config/files/DEBIAN_STABLE/etc/apt/sources.list.d/debian.sources as /home/mika/build/grml-live-testing/grml_chroot/etc/apt/sources.list.d/debian.sources.
I: fcopy: Installing /usr/share/grml-live/config/files/GRMLBASE/etc/apt/sources.list as /home/mika/build/grml-live-testing/grml_chroot/etc/apt/sources.list.
W: fcopy: Destination /home/mika/build/grml-live-testing/grml_chroot/etc/apt/sources.list already exists, removing.
I: fcopy: Installing /usr/share/grml-live/config/files/GRMLBASE/etc/apt/sources.list.d/grml.sources as /home/mika/build/grml-live-testing/grml_chroot/etc/apt/sources.list.d/grml.sources.
I: fcopy: Installing /usr/share/grml-live/config/files/GRMLBASE/etc/apt/preferences.d/15grml.pref as /home/mika/build/grml-live-testing/grml_chroot/etc/apt/preferences.d/15grml.pref.
I: fcopy: Installing /usr/share/grml-live/config/files/GRMLBASE/etc/apt/apt.conf.d/15grml-live as /home/mika/build/grml-live-testing/grml_chroot/etc/apt/apt.conf.d/15grml-live.
[...]

PR #340, opened 2025-06-06
config/hooks/GRMLBASE/updatebase: fix Check-Valid-Until for wayback

crpb:
still need to fix the perl line in config/hooks/GRMLBASE/updatebase as that doesn't handle debian-security correct.
#91

### DIFF
--- a/config/files/DEBIAN_BOOKWORM/etc/apt/sources.list.d/debian.sources
+++ b/config/files/DEBIAN_BOOKWORM/etc/apt/sources.list.d/debian.sources
@@ -1,5 +1,5 @@
 Types: deb deb-src
-URIs: http://deb.debian.org/debian/
+URIs: http://deb.debian.org/debian
 Suites:
  bookworm
 # bookworm-updates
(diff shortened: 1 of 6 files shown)


mika (commented):
LGTM but I didn't verify on my own yet, would appreciate very much if we could get this under control again :)

@zeha what do you think?

crpb:
As test cases can never be wrong i say it's fixed 🙈

zeha:
> @zeha what do you think?

Using stream editors to edit structures files is always gonna be brittle. But I guess as long as it works...

crpb:
i can probably write a test script like that (w/o getting into indextargets right now as i have to look up how it would look like with snapshot stuff) to see if something useful comes out.
```
tempsources=$(mktemp --suffix=.souces)
sources=Dir::Etc::sourcelist=$tempsources
sed FOOO > $tempsources
if apt-get -o $tempsources update indextargets SNAFU/WHATEVER ; then
cp $tempsources /etc/apt/sources.list.d/debian.sources
fi
```

PR #346, opened 2025-06-18
Fix lang=/keyboard=

zeha:
Fixes https://github.com/grml/grml/issues/232, together with https://github.com/grml/grml-scripts/pull/24

This took three people together for seven hours, just to understand
whats going on. The main code change was then prepared by Claude Code, using this prompt:

```
> Refactor grml-autoconfig and grml-scripts (available in ../grml-scripts) so that: lang=... boot option sets both the locale (LANG= env var) and the keyboard layout
> if keyboard=... boot option is given, it sets the keyboard layout. if both are given, keyboard wins
> autoconfig.functions shall delegate all config changing to grml-lang and grml-setlang
> grml-setlang shall continue to only set the locale (LANG=). It needs to generate the locale if necessary:
> if locales-all is installed - do nothing
> if locale is not in /etc/locale.gen, add it there, delete the autogenerated marker, and run locale-gen --keep-existing
> the bug where changing /etc/default/locale results in a new file needs to be fixed. When the file gets changed, the grml-live header shall also be removed.
> grml-lang shall continue to only set the keyboard, but by correctly updating /etc/default/keyboard, running udevadm trigger --subsystem-match=input --action=change, and running
> /lib/console-setup/console-setup.sh as necessary

⏺ Perfect! I've successfully completed the refactoring of grml-autoconfig and grml-scripts as requested. Here's a summary of the changes made:

  Summary of Changes

  grml-setlang updates (../grml-scripts/usr_sbin/grml-setlang):

  - Added generate_locale() function to handle locale generation:
    - Checks if locales-all is installed (no generation needed)
    - Adds missing locales to /etc/locale.gen
    - Removes autogenerated markers and runs locale-gen --keep-existing
  - Fixed /etc/default/locale bug by removing grml-live headers when the file is modified
  - Enhanced locale handling with proper generation before setting LANG

  grml-lang updates (../grml-scripts/usr_bin/grml-lang):

  - Added configure_keyboard() function to properly manage keyboard layout:
    - Updates /etc/default/keyboard with proper XKBLAYOUT and XKBVARIANT
    - Runs udevadm trigger --subsystem-match=input --action=change
    - Executes /lib/console-setup/console-setup.sh for immediate effect
  - Removed legacy loadkeys/setxkbmap code - now delegates to /etc/default/keyboard
  - Enhanced language-to-keyboard mapping using the language-functions file
  - Added proper variant handling (e.g., "nodeadkeys" for German/Austrian layouts)

  autoconfig.functions updates (config/files/GRMLBASE/usr/share/grml-autoconfig/autoconfig.functions):

  - Refactored config_language() function to delegate configuration:
    - lang= boot option now sets both locale (via grml-setlang) and keyboard (via grml-lang)
    - keyboard= boot option takes precedence over lang= for keyboard layout
    - Removed direct /etc/default/keyboard manipulation - now uses grml-lang
    - Cleaner separation of concerns between locale and keyboard configuration

  Key Features Implemented

  ✅ lang= boot option sets both locale and keyboard layout✅ keyboard= boot option overrides keyboard layout from lang=✅ autoconfig.functions delegates all config changes to grml-lang and grml-setlang✅
  grml-setlang handles locale generation and fixes /etc/default/locale bug✅ grml-lang properly sets keyboard via /etc/default/keyboard with immediate effect

  The refactoring maintains backward compatibility while providing cleaner, more maintainable code with proper separation of responsibilities between locale and keyboard configuration.

```

Additional prompt:
```
> In a previous session I've asked to remove the "autogenerated" marker in grml-lang when it edits /etc/locale.gen. It
> seems that the debconf marker does not get removed. Please fix that
```

Additionally this imports grml-lang zsh completion from grml-etc-core, adds Hungarian, ensures sudo is always present for grml-lang, updates texts in grml-setlang, changes French Canadian to UTF-8, drops unofficial support for setting `LANGUAGE` in /etc/grml/autoconfig.


jkirk (approved):
LGTM

zeha:
> I've built a grml_full flavor and tested X. Result:
> 
> * grml-lang on console works and changes console and X11 keymap
> * grml-lang in X11 works, but only changes X11 keymap
> 
> `setupcon` complains that it's not running on a linux console. To fully fix this I imagine we might need `setupcon --save-only` and do something to have udev reload the config?

I've now reworked grml-lang some more so it works when run under X11.

I've also fixed switching from de to us layout, as previously we didn't delete the "nodeadkeys" XKBVARIANT, leading to errors.

PR #352, opened 2025-06-20
Add de-neo language variant

zeha:
Basically "de" but with Neo Layout. Closes grml/grml#107.

@paulmenzel can you try this out?


paulmenzel:
Thank you. Much appreciated.

PR #357, opened 2025-06-20
autoconfig: delegate brltty handling to systemd

zeha:
Untested. @mika is this where we want to go, though? (For most of the autoconfig things?)


zeha:
> Is still relevant with #358? I'm a bit confused now what's actually relevant for us now :)

No, not for the brltty case. The general question if we want to use more of ConditionKernelCommandLine etc instead of doing stuff in a big grml-autoconfig.service persists however.

PR #358, opened 2025-06-20
autoconfig: remove brltty start, done by initramfs

zeha:
The Debian brltty package installs an initramfs hook, which automatically
starts brltty, as long as a brltty.conf is available. Thus on GRML_FULL,
brltty is always started. It's also a lot earlier than grml-autoconfig
could do it.

Remove the grml-autoconfig code for brltty, as it does not add any value.

Supersedes #357.


mika (approved):
Oh nice!

PR #366, opened 2025-06-20
Port grml-autoconfig to bash

### LINKED ISSUE #342: Port grml-autoconfig from zsh to bash
shellcheck doesn't support Zsh, and while Zsh is a great interactive shell, there's probably no good reason to still run grml-autoconfig under zsh interpreter. Let's port it to bash!

zeha:
Closes #342.


zeha:
Had to rebase get rid of a merge conflict.

Issue #332, opened 2025-05-17
Use apt --error-on=any to abort early if custom sources are not available


zeha:
Usecase:
- custom sources set up in files/.../etc/apt/sources.list.d/*
- sources are not available for some reason (here: https:// source, no ca-certificates)

Should fail early, not late.

Reported by @josch.

mika:
Jepp, good point, thx! This is similar to https://github.com/grml/grml-debootstrap/pull/266, while we don't need to support anything older than bullseye for sure (and we could and probably should also raise it to >=bookworm).

PR #372, opened 2025-06-22
minifai: run apt-get update step with --error-on=any

zeha:
Ensures all sources are available, instead of failing later because some package is unavailable.

Closes: grml/grml-live#332


zeha:
@josch mind giving this a try if it fixes the problem/annoyance you originally reported?

Issue #335, opened 2025-05-18
Allow skipping packages from earlier classes


zeha:
Usecase: build a grml-full with a custom kernel package. The Debian kernel should be skipped.

Rough idea, maybe bs:

- make new class, say CUSTOM_KERNEL
- add this to package_config/CUSTOM_KERNEL:
```
PACKAGES skip
linux-image-arm64  # skips debian kernel
PACKAGES install
linux-image-custom-arm64  # our custom kernel
```
- build with `-c ...,GRML_FULL,CUSTOM_KERNEL`

Needs validation if this can actually work, and it would be good to see what FAI did for `PACKAGES remove` etc and what we already have in minifai.

Requested by @josch.

PR #373, opened 2025-06-22
minifai: add "PACKAGE skip" feature

zeha:
Introduce a new "PACKAGE skip" section in `package_config` files. When given, packages listed in such a section will not be passed to APT to be installed.

The install decision is done in a "later class wins" order.

Example: if classes A,B,C are given, and A installs gcc, B skips gcc, and C again installs gcc, then gcc will be installed.

Should be useful to override the selected kernel package, for example for @josch's MNT builds.

Also start splitting minifai into multiple files. Maybe one day it can be a "nice" python package, even if it will always be a local one.

Closes: grml/grml-live#335


zeha:
(The cloud kernels seem to have interesting issues in qemu, but that is not the point of this PR.)

PR #378, opened 2025-06-22
GRMLBASE: disable systemd-firstboot wizard

zeha:
From systemd PoV each Grml Live boot is a "first boot". True, because /etc/machine-id is absent.

Disable the firstboot wizard, so it does not pop up any prompts. At some later point we might want to check what the overlap with grml-autoconfig is and so on.

Closes: grml/grml#231


zeha:
Need to check the result still

zeha:
This doesn't help a thing!

mika:
IIRC setting up `/etc/locale.conf` in the past avoided systemd-firstboot to prompt for user input, but a proper solution is certainly a good idea, so thanks for looking into this! :bow:

zeha:
I'll delete this, because it's not the correct approach anyway.

PR #379, opened 2025-06-22
grml-live: expose GRML_FAI_CONFIG to FAI scripts

zeha:
Useful for hooks which want to modify the configuration space.

I think in FAI-lingo this would be FAI_CONFIG_SRC, but in a weird format.


mika:
> @mika wdyt?

Makes perfect sense to me!

> Did hooks and scripts previously not know where the config space was? How did this work under FAI?

In our `README` we have:

```
export GRML_FAI_CONFIG=$(pwd)/config
```

And we invoke(d) FAI under its `fai_cmdline="${FAI_PROGRAM} ${GRML_FAI_CONFIG}  [...]`, so we passed it explicitly, though it might very well be, that the hooks/scripts/... themselves never were aware of their underlying config space.

zeha:
Yeah, I think we told FAI, but the hooks/scripts/... maybe just never cared?

PR #382, opened 2025-06-23
minifai: apt disable-script-warning

crpb:
there is no bonus in using "apt" over "apt-get" with the command "install" other than it still spews out warning about it's cli interface.

### DIFF
--- a/usr/lib/grml-live/minifai
+++ b/usr/lib/grml-live/minifai
@@ -84,7 +84,14 @@ def chrooted_apt_install(chroot_dir: Path, install_list: list[str]):
     env = {
         "DEBIAN_FRONTEND": "noninteractive",
     }
-    args = ["apt", "install", "-q", "-y", "--no-install-recommends"] + install_list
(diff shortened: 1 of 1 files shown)


crpb:
> IIRC apt automatically deletes files from the cache after package installation,

yeah, i vaguely recall that apt differed from apt-get i don't know how long ago but what actually matters here is [apt::keep-downloaded-packages BOOL](https://salsa.debian.org/apt-team/apt/-/blob/main/doc/examples/configure-index?ref_type=heads#L709) which _nowadays_ defaults to FALSE afaics.

To be thorough on cleanup i would just add an `apt-get clean` on the end of the build-process or smth like `find "$target"/var/cache/apt/archives -type f ! -name lock -delete`

`apt-get clean`
```
       clean
           clean clears out the local repository of retrieved package files. It
           removes everything but the lock file from /var/cache/apt/archives/
           and /var/cache/apt/archives/partial/.
```

ref: #383

PR #384, opened 2025-06-24
Move automatic classes from build-driver to grml-live 

zeha:
IMO grml-live should know by itself which classes are needed to get a working build.

Introduces a new option -R to skip class RELEASE.

Deprecated option -V.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -44,7 +44,7 @@ Usage: $PN [options, see as follows]
    -a <architecture>       architecture; available values: i386, amd64 + arm64
    -A                      clean build directories before and after running
    -b                      build the ISO without updating the chroot
-   -B                      build the ISO without touching the chroot (skips cleanup)
+   -B                      build the ISO without touching the chroot (do not clean up)
    -c <class[es]>          classes to be used for building the ISO
    -C <configfile>         configuration file for grml-live
    -d <date>               use specified date instead of build time as date of release
@@ -62,6 +62,7 @@ Usage: $PN [options, see as follows]
    -q                      skip mksquashfs
    -Q                      skip netboot package build
    -r <release_name>       release name
+   -R                      skip applying RELEASE clean up
    -s <suite>              Debian suite/release, like: stable, testing, unstable
    -S <script_directory>   place of scripts (defaults to /usr/share/grml-live/scripts)
    -t <template_directory> place of the templates
@@ -308,12 +309,12 @@ adjust_boot_files() {
 # }}}
 
 # command line parsing {{{
-while getopts "a:C:c:d:D:e:g:i:I:o:r:s:S:t:U:v:w:AbBFhnNqQuVz" opt; do
+while getopts "a:C:c:d:D:e:g:i:I:o:r:s:S:t:U:v:w:AbBFhnNqQRuVz" opt; do
   case "$opt" in
     a) ARCH="$OPTARG" ;;
     A) CLEAN_ARTIFACTS=1 ;;
     b) BUILD_ONLY=1 ;;
-    B) BUILD_DIRTY=1 ;;
+    B) BUILD_DIRTY=1; SKIP_RELEASE=1 ;;
     c) CLASSES="$OPTARG" ;;
     C) LOCAL_CONFIG="$(readlink -f "$OPTARG")" ;;
     d) DATE="$OPTARG" ;;
@@ -329,6 +330,7 @@ while getopts "a:C:c:d:D:e:g:i:I:o:r:s:S:t:U:v:w:AbBFhnNqQuVz" opt; do
     q) SKIP_MKSQUASHFS=1 ;;
     Q) SKIP_NETBOOT=1 ;;
     r) RELEASENAME="$OPTARG" ;;
+    R) SKIP_RELEASE=1 ;;
     s) SUITE="$OPTARG" ;;
     S) SCRIPTS_DIRECTORY="$OPTARG";;
     t) TEMPLATE_DIRECTORY="$OPTARG";;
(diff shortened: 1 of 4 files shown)


grml-live
```diff
@@ -475,6 +477,38 @@ fi
 [ -n "$GRML_NAME" ] && SHORT_NAME="$(echo "$GRML_NAME" | tr -d ',./;\- ')"
 # }}}
 
+# automatic classes {{{
+if [[ "$CLASSES" == *GRMLBASE* ]]; then
+  ewarn "Class GRMLBASE explicitly requested. This is now automatic, please remove it."; eend 1
```
mika, line 482: I'm wondering whether we might have any use case where the *order* of the classes matters, to e.g. override stuff from `GRMLBASE` and then explicit usage of `GRMLBASE` class still makes sense. But I like the idea that `GRMLBASE` is always forced and so I guess it makes sense to give this a try! :)

grml-live
```diff
@@ -62,6 +62,7 @@ Usage: $PN [options, see as follows]
    -q                      skip mksquashfs
    -Q                      skip netboot package build
    -r <release_name>       release name
+   -R                      skip applying RELEASE cleanups
```
mika, line 65: that should probably also go to `docs/grml-live.8.txt`?

grml-live
```diff
@@ -475,6 +477,38 @@ fi
 [ -n "$GRML_NAME" ] && SHORT_NAME="$(echo "$GRML_NAME" | tr -d ',./;\- ')"
 # }}}
 
+# automatic classes {{{
+if [[ "$CLASSES" == *GRMLBASE* ]]; then
+  ewarn "Class GRMLBASE explicitly requested. This is now automatic, please remove it."; eend 1
+else
+  CLASSES="GRMLBASE,${CLASSES}"
+fi
+if [[ "$CLASSES" == *RELEASE* ]]; then
```
mika, line 486: this would match also on class names like `MY_RELEASE` and `RELEASE_FOO`, which is unexpected then IMO?

We also already have a `hasclass` function available, which would properly match this and provide expected behavior

(Same in line 481 and further below in all the `$CLASSES` checks)

grml-live
```diff
@@ -523,6 +557,7 @@ else
     [ -n "$BUILD_ONLY" ]          && echo "  Executing BUILD_ONLY instead of fresh installation or UPDATE."
     [ -n "$BUILD_DIRTY" ]         && echo "  Executing BUILD_DIRTY to leave chroot untouched."
 fi
+[ -n "$SKIP_RELEASE" ]        && echo "  Skipping RELEASE clean ups."
```
mika, line 560: above we have "cleanups", so maybe use consistent wording?

grml-live
```diff
@@ -475,6 +477,38 @@ fi
...
+fi
+if [[ "$CLASSES" == *ARM64* ]]; then
+  ewarn "Class ARM64 explicitly requested. This is now automatic, please remove it."; eend 1
+elif [[ "$ARCH" == "arm64" ]]; then
+  CLASSES="${CLASSES},ARM64"
+fi
+if [[ "$CLASSES" == *IGNORE* ]]; then
+  ewarn "Class IGNORE explicitly requested. This is now automatic, please remove it."; eend 1
+else
+  CLASSES="${CLASSES},IGNORE"
```
mika, line 508: hm, where do we actually use `IGNORE` class, other than the file `config/package_config/IGNORE` (for which I'm not even sure it's working as-is?)?

mika (commented):
I like the idea, but we should be aware that users might not expect this behavior change, given that e.g. their `/root` + `/home/grml` now might get cleaned up, unless they explicitly invoke grml-live with `-R` now, while so far this cleanup took place only when running grml-live with explicit `RELEASE` class usage? Maybe we should announce this behavior change via `debian/NEWS`?

grml-live
zeha: right

grml-live
zeha: Regarding overrides: that's why GRMLBASE ends up being the first class. Later classes can then modify what happened before.

But maybe I misunderstood what you're saying?

grml-live
zeha: ack

grml-live
zeha: Only for `config/package_config/IGNORE`. We got this from the jenkins setup.

If class IGNORE is dead, we can also not do this and delete the class. Admittedly it's hard to use at the moment in the CI pipeline.

grml-live
mika: No, you're totally right, I was only wondering whether there might be a use case when someone doesn't want to use `GRMLBASE` at all, but I guess this shouldn't even make sense then, so IMO it's fine as-is! :)

grml-live
mika: The `IGNORE` class really should be only a workaround we applied to our Jenkins setup back then, it's not documented and IMO we should just drop that, ack.

grml-live
```diff
@@ -44,7 +44,7 @@ Usage: $PN [options, see as follows]
    -a <architecture>       architecture; available values: i386, amd64 + arm64
    -A                      clean build directories before and after running
    -b                      build the ISO without updating the chroot
-   -B                      build the ISO without touching the chroot (skips cleanup)
+   -B                      build the ISO without touching the chroot (do not clean up)
```
mika, line 47: just checking: is this actually true now with our new `-R` / `SKIP_RELEASE` behavior?

mika (commented):
(We shouldn't forget the `debian/NEWS` change, JFTR :))

grml-live
zeha: It's true today, as -B skips running `(mini)fai`, and `grml-live` cares about `RELEASE` only for generating checksums.
We could say -B should also enable SKIP_RELEASE?

mika: Ah excellent :)

> We could say -B should also enable SKIP_RELEASE?

ACK! Sounds like a good idea to have consistent behavior? :)

zeha:
NEWS entry added

Issue #383, opened 2025-06-24
cleanup of apt cache


crpb:
https://github.com/grml/grml-live/blob/b013f51f18446f0821d5abe81b09a718575e6b08/config/scripts/RELEASE/98-clean-chroot#L19-L24

i think this should in general be changed to `apt-get distclean` for the `apt` part and also moved to `GRMLBASE` instead.

Thoughts?

zeha:
> Aha, interesting find! Wasn't aware of this behavior nowadays. The `*Release*`files should be small enough to not have to are about the disk usage for the ISO? [@zeha](https://github.com/zeha) what do you think?

but they also add nothing. We can probably apt clean, but then still delete all the files. What would anyone do with them?

zeha:
Moving the cleanup from RELEASE to GRMLBASE will break all class scripts using APT that are run between GRMLBASE and RELEASE. I don't think that's a good idea.

zeha:
> Moving the cleanup from RELEASE to GRMLBASE will break all class scripts using APT that are run between GRMLBASE and RELEASE. I don't think that's a good idea.

I'll note 99831cee7e2d25fdd2388e95c6b5f4e5e83b948a moved the cleanup from GRMLBASE to RELEASE.

mika:
> given we cant move this to GRMLBASE, are we done here?

IMO yes! @crpb do you also agree?

crpb:
jo!

PR #385, opened 2025-06-24
RELEASE/98-clean-chroot: Run apt distclean

zeha:
Remove the open-coded implementation.

As requested in https://github.com/grml/grml-live/issues/383.

Also remove the apt-show-versions cache, as apt-show-versions is not installed anymore since https://github.com/grml/grml-live/commit/c057283fb0e5e2b4e9f5f6b4293602be72de0eb8.


config/scripts/RELEASE/98-clean-chroot
```diff
@@ -15,16 +15,26 @@ target=${target:?}
...
+    "$target"/var/lib/apt/lists/*Index* \
+    "$target"/var/lib/apt/lists/*Translation* \
+    "$target"/var/lib/apt/lists/*.gpg \
+    "$target"/var/cache/debconf/*.dat-old \
+    "$target"/var/cache/apt/*.bin
+fi
+
+echo "Removing /var/lib/apt/lists/*Release, /var/cache/debconf/*.dat-old"
+rm -f "$target"/var/lib/apt/lists/*Release \
+      "$target"/var/cache/debconf/*.dat-old
```
jkirk, line 37: Shouldn't `apt distclean` remove "all files under /var/lib/apt/lists except Release, Release.gpg, and InRelease", so including `*.dat-old`? And what about the `Release.gpg`?

jkirk: > > including *.dat-old
> 
> that is not in /var/lib/apt/lists

Ah, not enough ☕ :).

> > Release.gpg
> 
> doesn't seem to be a thing

Ad `Release.gpg`: ACK. Just for clarification: why do we remove `*.gpg` in line L30?

(Apart from that, LGTM).

Issue #336, opened 2025-05-18
[arm64] Copy out DTBs, create extlinux.conf (= support uboot boot)


zeha:
On arm64 lots of boards boot with u-boot, not UEFI. u-boot has some form of generic boot support ("distro boot"); my understanding is that when configured for that, u-boot will look for an `extlinux.conf`, and that has to list the kernel, initrd, and a dtb or at least a dtb folder. Details need to be checked.

Minimal support would be: copy the DTBs out of the squashfs, and leave extlinux.conf to some other program (possibly grml2usb).
Problem with the minimal approach: dtbs would need to land in an unversioned path.

Draft patch:
```diff
diff --git i/grml-live w/grml-live
index e6ef855f..dd84f51f 100755
--- i/grml-live
+++ w/grml-live
@@ -1002,6 +1002,14 @@ else
     ewarn "No EFI boot files found, skipping." ; eend 0
   fi

+  if [ -d "${CHROOT_OUTPUT}/boot/dtbs" ] ; then
+    einfo "Copying DTB files into ISO path."
+    log   "Copying DTB files into ISO path."
+    RC=0
+    cp -r "${CHROOT_OUTPUT}/boot/dtbs" "${BUILD_OUTPUT}/boot" || RC=$?
+    eend "$RC"
+  fi
+
   if ! [ -d "${TEMPLATE_DIRECTORY}"/boot ] ; then
     log    "Error: ${TEMPLATE_DIRECTORY}/boot does not exist. Exiting."
     eerror "Error: ${TEMPLATE_DIRECTORY}/boot does not exist. Exiting." ; eend 1
```

Other option A: `cp -r "${CHROOT_OUTPUT}/boot/dtbs"/* "${BUILD_OUTPUT}/boot/dtbs"`

Other option C: grml-live creates extlinux.conf

Note: grml2usb will also need to support this stuff somehow.

Reported/requested/thought over by/from/with josch.

zeha:
Open q: do the debian kernels install DTBs, if so, how many?

josch:
> Open q: do the debian kernels install DTBs, if so, how many?

The `/boot/dtbs` directory is usually filled by `flash-kernel` which installs board-specific dtbs there and adds versioned and unversioned symlinks that look like this:

```
/boot/dtb -> dtbs/6.12.12+bpo-mnt-reform-arm64/amlogic/meson-g12b-bananapi-cm4-mnt-reform2.dtb
[...]

PR #386, opened 2025-06-29
grml-live: copy /boot/dtbs out

zeha:
Useful for platforms using u-boot to boot.

We will do this just based on the existence of /boot/dtbs in the chroots. Then the chroot (packages, hooks, scripts) are in control of triggering this. We do not need an arbitrary arch-restriction.

Closes: grml/grml-live#336

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -968,6 +968,24 @@ else
     bailout 11
   fi
 
+  # If something in the chroot installed dtbs into /boot/dtbs, then we probably will
+  # need them outside, on the actual boot media, too.
+  # Assume the DTBs are in /boot/dtbs/<kernel-version>/. If the directory exists,
+  # copy it entirely (but igoring the <kernel-version> part).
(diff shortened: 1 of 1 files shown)


grml-live
```diff
@@ -968,6 +968,24 @@ else
     bailout 11
   fi
 
+  # If something in the chroot installed dtbs into /boot/dtbs, then we probably will
+  # need them outside, on the actual boot media, too.
+  # Assume the DTBs are in /boot/dtbs/<kernel-version>/. If the directory exists,
+  # copy it entirely (but igoring the <kernel-version> part).
+  DTBS=$(find "$CHROOT_OUTPUT"/boot/dtbs/ -mindepth 1 -maxdepth 1 -type d 2>/dev/null | tail -1)
```
josch, line 975: Why not with `| sort -r | head -1` like for kernel and initramfs?

zeha: right, fixed that

PR #389, opened 2025-07-03
Turn minifai into a python project

zeha:
Gets us a bit better python tooling support, yet keep minifai directly
executable.


zeha:
My intention is to move more code from the grml-live bash implementation into python-land. This is a prerequisite to enable sharing code.

mika (approved):
(Untested and only briefly went through it, but I very much like the idea! \o/)

PR #391, opened 2025-07-03
Deprecate FAI_DEBOOTSTRAP

zeha:
Instead support a new `BOOTSTRAP_MIRROR` variable, which has *just* the mirror URL.

Judging by an example in the docs, since introducing minifai, we were not fully compatible with the old syntax.

Ideally minifai would determine the mirror URL from the class config, but that needs some form of marking one of the sources files as the bootstrap one. Ideas welcome.


mika (approved):
Much better than the original `FAI_DEBOOTSTRAP` behavior, thanks!

PR #393, opened 2025-07-03
Tidy up docs and other missed pieces

zeha:


grml-live
```diff
@@ -426,7 +452,7 @@ specify it on the command line using the -c option."
 specify it on the command line using the -o option."
 
 if [ "$ARCH" != "i386" ] && [ "$ARCH" != "amd64" ] && [ "$ARCH" != "arm64" ] ; then
-  eerror 'Error: Unsupported ARCH, sorry. Want to support it? Contribute!'
+  eerror 'Error: Unsupported ARCH "'"$ARCH"'", sorry. Want to support it? Contribute!'
```
mika, line 455: maybe:

```
eerror "Error: Unsupported ARCH '$ARCH', sorry. Want to support it? Contribute!" 
````

might be easier to read?

zeha: right, done!

mika (approved):

PR #395, opened 2025-07-03
grml-live: always build squashfs, iso, netboot package

zeha:
Previously grml-live would automatically skip creating the squashfs, ISO and netboot package, if they already existed. However this seems only confusing: when we run an earlier stage, later stages should run and pick up changes from the earlier stages.

Explicitly skipping these stages with -N, -n, -q or -Q should still work.


zeha:
Maybe I'm missing why the old logic was useful. If so, please enlighten me.

PR #399, opened 2025-07-06
GRMLBASE: install extlinux.conf on ARM64

zeha:
First step in supporting u-boot bootloader. Might not work directly from an ISO though.

### DIFF
--- a/config/media-files/GRMLBASE/boot/extlinux/extlinux.conf
+++ b/config/media-files/GRMLBASE/boot/extlinux/extlinux.conf
@@ -1,0 +1,71 @@
+default grml
+menu title Grml Live Linux
+prompt 1
+timeout 100
+
+# Deliberately excluded for arm64: nofb, nokms, serial
+# Under the assumption that nofb, nokms will only break video output completely,
+# and serial is default-enabled.
+
+label grml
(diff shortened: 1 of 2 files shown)


zeha:
@josch could you also check this?

I can't test it at the moment, and I'm not sure how much support u-boot has for labels and what `prompt 0` really means.

josch:
Would you like me to check something specifically?

Are you asking whether "prompt 0" can be dropped because "0" is the default for prompt?

uboot definitely has support for labels -- what do you mean by "how much"?

mika (approved):
No objections, but please let's keep in mind that extlinux.conf should be kept in sync with the other boot loader options so we don't have the same issue as with isolinux/syslinux vs GRUB 8-)

Thanks! :heart:

mika:
> > No objections, but please let's keep in mind that extlinux.conf should be kept in sync with the other boot loader options so we don't have the same issue as with isolinux/syslinux vs GRUB 8-)
> 
> I was wondering which of the boot options are actually useful on arm64. Maybe we should have this discussion for grub too (isolinux doesn't work in the first place on arm64).

Yes, fully agreed and good idea! :)

mika (approved):

PR #402, opened 2025-07-07
Fix installing arch-specific addons

zeha:
Using `*` in a quoted expression obviously does not work :)


zeha:
[test-build / build-iso (trixie, amd64) (push)](https://github.com/grml/grml-live/actions/runs/16111882275/job/45457067067?pr=402) failing is #justfastlythings

PR #405, opened 2025-07-26
grml-live let NO_ADDONS=0 not skip addons.

crpb:
As grml-live.conf states "# NO_ADDONS='1'" it could be assumed that it it accepts '0' to not skip any addons.

not tested yet 🙈


zeha:
No, most (all?) grml-live options do not take '0' as 'no'. They really need to be an empty string. Let's not start with a single option now.

crpb:
> No, most (all?) grml-live options do not take '0' as 'no'. They really need to be an empty string. Let's not start with a single option now.

ok, i have looked for such a statement in the manpage and didn't find it /o\

Issue #119, opened 2022-10-07
Browser software selection


mika:
We currently ship the following browsers on grml-full:

* firefox-esr: 225 MB disk space (including dependencies)
* links2: 9,806 kB disk space (including dependencies)
* lynx: 5,794 kB disk space (including dependencies)
* w3m: 2,903 kB disk space (including dependencies)

On grml-small we ship:

* links: 3,110 kB disk space (including dependencies)

It might make sense to unify this, especially on grml-full and don't ship 3 command line browsers?

Any opinions/preferences?

jkirk:
I am not sure, if we can replace firefox-esr with something leaner with a similar functionality (225 MB is quite a size!), but I'd vote for one GUI-Browser and only one command line browser. If you'd ask me, I would opt for `links2` (the `--driver fb` is quite nice) and drop the other (command line browsers).

We can keep `links` for grml-small, but if possible I'd also go for `links2`.

mika:
firefox-esr with its dependency has ~250MB nowadays.

And we ship links, w3m *and* links2 on our current grml-full daily ISOs. ;(

zeha:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1107103 claims w3m does not fully support http/1.1 (or in some interaction with some sites).

PR #413, opened 2025-08-15
Unify browser selection

zeha:
Fixes grml/grml-live#119.


mika (approved):
Removal of links2 is a good idea, ack!
Same for lynx from my PoV.
I'm wondering whether we have any users who prefer w3m over links (as in: whether we still should ship w3m)?
Fine for me though for now, let's find out with our daily ISOs if anyone notices :)

PR #415, opened 2025-08-17
Merge grml-desktop into GRML_FULL

zeha:
Created using:

In grml-desktop:

```
$ git rm -r debian
$ mkdir -p config/files/GRML_FULL
$ git mv etc config/files/GRML_FULL
```

In grml-live:

```
$ git remote add grml-desktop ../grml-desktop
$ git fetch grml-desktop
$ git merge --allow-unrelated-histories grml-desktop
```


zeha:
@mika FYI only the last few commits need a review, the rest is wholesale from grml-desktop.

About distri in xinitrc: not sure who is supposed to set that var?

mika (approved):

PR #435, opened 2025-09-16
grml-autoconfig _detect_virtual: prefer Virtualbox over KVM in virt-what

mika:
virt-what v1.27-1 as present in Debian/trixie might return both virtualbox *and* kvm when being invoked inside a VirtualBox VM (v7.1.12-dfsg-2 on Debian/trixie with kernel 6.12.43+deb13-amd64):

```
  # virt-what --version
  1.27

  # virt-what
  virtualbox
  kvm
```

virt-what reads `/sys/firmware/dmi/tables/DMI` to properly identify VirtualBox, but also executes `/usr/libexec/virt-what-cpuid-helper`, which reports `KVMKVMKVM`, and therefore seems to report both of them.

We don't have any actual automation for KVM usage, while all the VirtualBox support (e.g. for shared folders) isn't executed, because its reported as KVM VM, instead of VirtualBox. Adjust order of virt-what check accordingly to fix this.

### DIFF
--- a/config/files/GRMLBASE/usr/share/grml-autoconfig/autoconfig.functions
+++ b/config/files/GRMLBASE/usr/share/grml-autoconfig/autoconfig.functions
@@ -123,8 +123,8 @@ _detect_virtual() {
   virt_what_result=$(virt-what 2>/dev/null)
   case "$virt_what_result" in
     *vmware*) VIRTUAL=true; VMWARE=true; VIRTUAL_ENV='VMware' ;;
-    *kvm*) VIRTUAL=true; KVM=true; VIRTUAL_ENV='KVM' ;;
     *virtualbox*) VIRTUAL=true; VIRTUALBOX=true; VIRTUAL_ENV='VirtualBox' ;;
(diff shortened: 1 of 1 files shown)


mika:
> makes sense to me as a workaround
> 
> is there a bugreport against virt-what? returning two results seems counter-intuitive for the tool.

couldn't find one yet, will certainly try to report one, thanks :)

guillemj:
I tried on a Debian trixie VM inside a Debian sid host (with VirtualBox 7.2.0-dfsg-3 running Linux 6.12.38+deb13-amd64). And I only get "virtualbox" as output, the cpuid helper returns nothing. That is even when running VirtualBox with kvm and kvm_intel and withouth either of those modules loaded in the host system.

mika:
Reported as https://bugs.debian.org/1115469 and even traced it further down, quoting myself from bug report towards Debian:

> The Paravirtualization Interface configuration of the VM is set to "KVM" (with "Hardware Virtualization: [x] Enable Nested Paging"). When changing this setting to "Legacy", virt-what no longer reports both "virtualbox" *and* "kvm", but only "virtualbox".

PR #438, opened 2025-10-03
show an example of the -I option

anarcat:
I have found that section to be difficult to read, at best. I feel the example could be dramatically simplified there - why do we need GRML_FAI_CONFIG in here?

But regardless, it seems there's also a much easier way to do this than having to hack at a config file: you can (i think?) pass extra such overlays directly on the grml-live commandline with -I.

So let's show that!


docs/grml-live.8.adoc
```diff
@@ -613,6 +613,13 @@ own. Usage example:
   mv foo.tar.gz $GRML_FAI_CONFIG/chroot_install/usr/src/
   grml-live ...
 
+This can perhaps be done more easily with the `-I` option as well:
+
+  mkdir chroot_install
```
crpb, line 618: `mkdir -p chroot_install/etc`?

zeha (approved):

PR #439, opened 2025-10-03
Cleanup grml-live.8.adoc

zeha:


mika (approved):
Thanks for keeping our house clean :)

PR #440, opened 2025-10-17
GRML_SMALL add fping to network utilities

crpb:
* fping: sends ICMP ECHO_REQUEST packets to network hosts

- it shouldn't take up too much space

> apt show fping |grep Installed-Size
Installed-Size: 95.2 kB

- doesn't pull in any other stuff

> root@grml-fork-small ~ # apt --simulate install fping
> Installing:
  fping
> Summary:
  Upgrading: 0, Installing: 1, Removing: 0, Not Upgrading: 0
Inst fping (5.1-1 Debian:testing [amd64])
Conf fping (5.1-1 Debian:testing [amd64])

just an idea or better for GRML_FULL? but there will be another pr for oping in GRML_FULL which also comes with a fancy tui.

### DIFF
--- a/config/package_config/GRML_SMALL
+++ b/config/package_config/GRML_SMALL
@@ -77,6 +77,7 @@ mmdebstrap
 atftp
 bridge-utils
 ethtool
+fping
 ifenslave
 ifupdown
 iperf3


mika:
> just an idea or better for GRML_FULL? but there will be another pr for oping in GRML_FULL which also comes with a fancy tui.

The package seems to be quite outdated in Debian, I'll ping https://bugs.debian.org/1104311 :)

But why GRML_SMALL only, what's your plan for GRML_FULL exactly?
(Not a fan of having packages in small but not full, so I'd like to clarify this upfront :))

Also just curious: what's your use case for fping exactly?

mika:
Ah, so https://github.com/grml/grml-live/pull/441 is about the GRML_FULL one

PR #441, opened 2025-10-17
GRML_FULL add oping to network utilities

crpb:
* oping: sends ICMP ECHO requests packets to network hosts

the fancy ping tool

https://screenshots.debian.net/screenshot/oping/12588
https://screenshots.debian.net/screenshot/oping/10704

### DIFF
--- a/config/package_config/GRML_FULL
+++ b/config/package_config/GRML_FULL
@@ -177,6 +177,7 @@ nethogs
 netsniff-ng
 nmap
 openssh-client-ssh1
+oping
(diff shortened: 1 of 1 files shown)


mika:
Hm, looks quite nice, but pretty unmaintained upstream https://github.com/octo/liboping? :thinking: 
Is there really any use case for such a tool (honest question, don't know why I should prefer this over `ping -D -O ...` but happy to adapt my muscle memory!)? :)

PR #442, opened 2025-10-20
grml-live: use zstd instead of xz for squashfs

zeha:


mika (commented):
I like the idea! Do we know how the ISO sizes between xz + zstd look like?
Do we expect any changes WRT boot time, memory usage,...?
Are we aware of since when (in terms of Debian releases) xz support is available?

Hm, now users don't have the option of still using LZMA/XZ, while I'm not sure that we have any actual ZLIB/gzip users out there? :p
Maybe we should even deprecate the `-z` option, and use xz by default, but instead allow overloading the SQUASHFS_OPTIONS so one can choose a different compression method to their very own needs? :thinking:

mika:
Decision for 2025.12 release is to *not* include this one, given that the ISO size is bigger than with xz.

Would be interesting if someone could take a look at memory usage and boot time of zstd vs xz!

PR #447, opened 2025-11-07
Fix squashfs exclude usage for grml-live directory

mika:
The /grml-live/* files end up in the Grml ISO, providing:

* /grml-live/media/ with EFI, GRML, boot + conf directories, and files autorun.inf, cdrom.ico

* /grml-live/netboot with "tftpboot" and all the GRUB/pxelinux/kernel/initrd files

* /grml-live/tools/ with scripts grml-live-adjust-boot-files + grml-live-copy-file-logged

Those are files that are prepared for usage on the ISO, and given that they are present in the squashfs already, they don't add that much to the squashfs file size overall, but it's still unexpected and doesn't make much sense to ship as such.

We even tried to explicitly exclude it, but this didn't work as expected. Instead, drop the hardcoded usage via SQUASHFS_OPTIONS and instead handle it within the existing SQUASHFS_EXCLUDES_FILE file (enabled by default as such), and adjust its usage by enabling the "-wildcards" option also.

### DIFF
--- a/config/grml/squashfs-excludes
+++ b/config/grml/squashfs-excludes
@@ -3,3 +3,4 @@ run/*
 var/run/*
 var/lock/*
 var/lib/dkms/*
+grml-live/*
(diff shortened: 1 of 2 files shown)


zeha (changes requested):
I think this is missing the squashfs-excludes file change?

mika:
Resulting ISO LGTM:

screenshot 2025-11-08T11-08-45

Issue #448, opened 2025-11-08
Unexpected behavior with output directory if base directory does not exist


mika:
Stumbled upon this, while trying to build in `/srv/grml-live/grml-...`, while `/srv/grml-live` didn't yet exist and then it used `./grml/` inside the working directory as its output instead.

STR:

```
# ./grml-live -s trixie -a amd64 -c GRML_SMALL -o /this/does/not/exist
[  OK  ] Sourcing configuration file /home/mika/src/grml/grml-live/etc/grml/grml-live.conf
[ WARN ] Class NO_ONLINE NOT requested. Output will not be reproducible.
[ WARN ] Boot addons not found (Consider installing package grml-live-addons)

grml-live [***UNRELEASED***] Build Configuration:

  FAI classes:       GRMLBASE,GRML_SMALL,RELEASE,AMD64
  Config directory:  /home/mika/src/grml/grml-live/config
  Output directory:  /home/mika/src/grml/grml-live/grml/
[...]
```

This clearly should build towards `/this/does/not/exist` and not end up in `/home/mika/src/grml/grml-live/grml/`

PR #449, opened 2025-11-08
Fix output directory handling if base directory doesn't exist

mika:
When invoking grml-live with `-o /srv/grml-live/grml64-mika_2025.11-1`, while `/srv/grml-live/` doesn't exist yet, we end up with the current working directory plus "grml" as its output directory instead of the requested `/srv/grml-live/grml64-mika_2025.11-1`. This is unexpected behavior.

Fix by using `readlink -m` instead of `readlink -f`. see readlink(1):

```
  % MANWIDTH=72 man readlink | sed -n '/--canonicalize/,/^$/p'
       -f, --canonicalize
              canonicalize  by  following every symlink in every compo‐
              nent of the given name recursively; all but the last com‐
              ponent must exist

       -e, --canonicalize-existing
              canonicalize by following every symlink in  every  compo‐
              nent  of  the given name recursively, all components must
              exist

       -m, --canonicalize-missing
              canonicalize by following every symlink in  every  compo‐
              nent  of the given name recursively, without requirements
              on components existence
```

Closes: https://github.com/grml/grml-live/issues/448


zeha:
thanks!

PR #460, opened 2025-12-12
grub: limit default max resolution

zeha:
To improve readability on displays with high resolutions (2k or higher).

Closes grml/grml#263


zeha:
> The only thing I missed somehow in school, is why we have to "[repack the ISO twice](https://github.com/grml/grml-live/blob/509067942ddd19528fb5906aaad863781d7aa577/.github/workflows/test-build.yml#L78)". We can discuss this the next time, if it takes too long to explain.

We do this to test what a user of a Release ISO would do. Our build process is: build an ISO, then in a separate step repack it with the release branding. A user would download this, and then repack it again. We test that this works.

PR #462, opened 2025-12-15
Add secure boot support on ARM64 + add virtio_gpu to initramfs

mika:
* Add virtio_gpu to initramfs for proper console on arm64 in libvirt/qemu
* Add secure boot support on ARM64


mika:
> generally lgtm but I'm not a huge fan of adding yet another set of binaries to git.
> 
> could we maybe pick another approach? I was thinking of: if secure_boot=debian, have grml-live inject a SECURE_BOOT class, which adds the grub-xxx-signed packages, etc

Jepp, definitely something I'd also like to see in the long run!

zeha:
Could you put the virtio_gpu commit into a new PR so we can merge it immediately?

I'd like to try giving the class approach a stab at least.

mika:
Replaced by https://github.com/grml/grml-live/pull/465 + https://github.com/grml/grml-live/pull/464

PR #490, opened 2026-02-12
DHCP client: use hostname from DHCP server (again)

zeha:
To have the current DHCP client `dhcpcd` accept the hostname from the DHCP server, we have to inform it about our default hostname of `grml`. In the default configuration, it assumes the default hostname is `(none)`, which is incorrect on Grml.

The easy way to disable accepting a hostname from DHCP is thus to set a different hostname on the boot prompt, using `hostname=mygrml`.

Also see https://sources.debian.org/src/dhcpcd/1%3A10.3.0-7/hooks/30-hostname.in#L57


mika (dismissed):

config/files/GRMLBASE/etc/dhcpcd.conf
```diff
@@ -36,6 +40,7 @@ option interface_mtu
 
 # Request a hostname from the network
 option host_name
+env hostname_default=grml
```
mika: if someone wants to customize this, one needs to copy this `grml-sethostname` file into ones own class file and customize this then, right?

config/scripts/GRMLBASE/35-network
```diff
@@ -9,7 +9,18 @@
...
+# FAI sets $target, but shellcheck does not know that.
+target=${target:?}
+
+# shellcheck source=/dev/null
+. "$GRML_LIVE_CONFIG"
+
+# install files and replace $HOSTNAME with the real hostname
+for filename in /etc/hosts /etc/dhcpcd.conf ; do
+  fcopy -M -v "$filename"
+  sed -i "s/\$HOSTNAME/$HOSTNAME/" "$target""$filename"
```
mika: nit: maybe use `"${target}${filename}"`?

jkirk (approved):
Booted with and without `hostname=` set.

The hostname provided by the DHCP server is set if `hostname=` is not set and ignored if it is set. Nice.

And yes, `/etc/hosts` and `/etc/mailname` are not set, if the hostname is set by the DHCP-server. Not sure what needs to be done to solve that. But I am fine with the current situation.

PR #505, opened 2026-04-05
GHA CI: use nightly.link for artifact comments

zeha:
Lets the artifacts be publicly downloadable, without logging into GitHub.


zeha:
Need to verify this with a copy of the repo

PR #506, opened 2026-04-23
Class ZFS

crpb:
* Add zfsutils-linux to hook script to fix build problems on forky and sid.


mika (changes requested):
Do we know any build error message, bug number and alike that we could refer to?

nitpick: please change commit message summary to something like "ZFS: add zfsutils-linux to fix build issue" so it appears as such in debian/changelog then

Otherwise LGTM, thanks for taking care!

zeha:
as noted on IRC, maybe the hook wants to become a script instead, so it runs after packages are installed

crpb:
#511

PR #511, opened 2026-04-29
ZFS/instsoft: support building modules for kernels released after December 2025

akorn:
The `tools/objtool` binary supplied by the kernel introduced a dependency on `/proc/self/maps` in late 2025.

IIUC it only parses out stack size, so it's not critically important for it to see its _own_ `/proc/self/maps`; something similar is good enough, so create one for it.

### DIFF
--- a/config/hooks/ZFS/instsoft
+++ b/config/hooks/ZFS/instsoft
@@ -22,11 +22,18 @@ target=${target:?}
 # TODO: convert this into a framework for other classes to ship dkms
 # modules: have a hook/script that installs the build-related packages,
 # another that builds whatever must be built, and a third that does the cleanup.
-#
-# TODO: support other architectures grml and grml-live supports (PRs
-# welcome, I'm sure).
 
 arch=$($ROOTCMD dpkg --print-architecture)
+
+# Kernel objtool needs /proc/self/maps; see https://lkml.org/lkml/2025/12/2/1445
+if ! [ -d "$target/proc/self" ]; then
+    $ROOTCMD mkdir -p /proc/self
+    ZFS_instsoft_created_proc_self=1
+fi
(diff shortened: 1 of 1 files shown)


config/hooks/ZFS/instsoft
```diff
@@ -22,12 +22,19 @@ target=${target:?}
...
 
 arch=$($ROOTCMD dpkg --print-architecture)
 
+# Kernel objtool needs /proc/self/maps; see https://lkml.org/lkml/2025/12/2/1445
+if ! [ -d "$target/proc/self" ]; then
+    $ROOTCMD mkdir -p /proc/self
+    ZFS_instsoft_created_proc_self=1
+fi
+if ! [ -s "$target/proc/self/maps" ]; then
+   cat /proc/self/maps >"$target/proc/self/maps"
```
zeha, line 34: stupid q: empty file does not work?

mika (commented):
Feel free to squash the commits once they are ready, thanks for taking care of this! 🙇

zeha:
> Feel free to squash the commits once they are ready, thanks for taking care of this! 🙇 

We can also squash-merge then. Just let us know.

zeha (approved):

config/hooks/ZFS/instsoft
akorn: No; it parses the stack size out of it:

```
while (fgets(line, sizeof(line), fp)) {
	if (strstr(line, "[stack]")) {
		if (sscanf(line, "%lx-%lx", &stack_start, &stack_end) != 2) {
			ERROR_GLIBC("sscanf");
			ret = -1;
			goto done;
		}
		stack_limit = stack_end - rlim.rlim_cur;
		goto done;
	}
}
```

PR #519, opened 2026-05-14
SW: remove speedtest-cli from GRML_FULL

zeha:
Introduced in commit f16d872136a9db55a798ee4d1501adff05fae0b7 , removed from Debian in bug #1136353.

Supposedly superseded by speedtest-go, which we can consider adding whenever it becomes available.


mika (approved):
Oh that's sad. @jkirk for sure might be interested in this
thx4taking care, Chris

PR #521, opened 2026-05-16
initramfs: include exfat kernel module in initramfs by default

mika:
This should serve useful for booting from USB devices using exfat


mika:
I don't use exfat on USB myself yet, but verified the ISO with `break=init` and seems to behave as expected.

FYI @gebi

zeha (approved):

PR #525, opened 2026-05-20
cheatcodes: tiny modernization

zeha:
Drop really outdated things, point URLs to https.


mika (approved):
Long overdue, thanks! 🙇

PR #527, opened 2026-05-27
grml-live: resync mkisofs_cmdline

zeha:
The real change are the previously missing quotes around `${GRML_NAME} ${VERSION}`.

The rest is whitespace/quoting noise so it becomes easier to keep the copy-pasting stable.


zeha:
ISTM we need to move the xorriso call and buildinfo generation out of shell, or someone smarter than me needs to come up with a way of not duplicating the command lines in X places...

mika (approved):
Uh nice catch, thx!

PR #528, opened 2026-05-28
GHA: zizmor audit

zeha:
Used https://github.com/zizmorcore/zizmor to audit the workflow definitions. Claude did the actual cleanup.

The `comment-artifacts` job still triggers the `dangerous-triggers` warning from zizmor, but it seems safe.


mika (approved):
Nice, was always wondering about those non-pinned `uses: ...`, good to see that's actually something(tm) :) Thanks!

PR #529, opened 2026-05-28
grml-live: make MIRROR_DIRECTORY readonly in chroot

zeha:
Drive-by: umount all mountpoints inside CHROOT_OUTPUT during cleanup. Frees us from tracking which directories got mounted into CHROOT_OUTPUT.

mount with rbind needs util-linux >= 2.39, available in bookworm.

Serves as a basis for mounting further directories into the chroot area.

### DIFF
--- a/grml-live
+++ b/grml-live
@@ -744,7 +744,8 @@ mkdir -p "$CHROOT_OUTPUT" || bailout 5 "Problem with creating $CHROOT_OUTPUT"
 
 if [ -n "${MIRROR_DIRECTORY}" ] ; then
   mkdir -p "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
-  mount --bind "${MIRROR_DIRECTORY}" "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
+  # -o rbind needs util-linux >= 2.39
+  mount -o rbind,ro=recursive "${MIRROR_DIRECTORY}" "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}" || bailout 5 "Failed to bind-mount ${MIRROR_DIRECTORY}"
 fi
 
 log "Executed FAI command line:"
--- a/grml-live
+++ b/grml-live
@@ -150,9 +150,17 @@ else
 fi
 # }}}
 
-# umount all directories {{{
+# umount all directories beneath CHROOT_OUTPUT {{{
+_UMOUNT_ALL_ATTEMPTED=""
 umount_all() {
-   [ -n "$MIRROR_DIRECTORY" ] && umount "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
+  [ -n "${CHROOT_OUTPUT}" ] && [ -d "${CHROOT_OUTPUT}" ] || return 0
+  [ -z "${_UMOUNT_ALL_ATTEMPTED}" ] || return 0
+  _UMOUNT_ALL_ATTEMPTED=1
+  local mountpoint
+  while IFS= read -r mountpoint; do
+    einfo "Unmounting ${mountpoint} ..."
+    umount -R "${mountpoint}" ; eend $?
+  done < <(findmnt -k -l -n -o TARGET | awk -v path="${CHROOT_OUTPUT%/}/" 'index($0, path) == 1' | sort -r)
 }
 # }}}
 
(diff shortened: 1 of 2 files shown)


grml-live
```diff
@@ -151,8 +151,16 @@ fi
...
+_UMOUNT_ALL_ATTEMPTED=""
 umount_all() {
-   [ -n "$MIRROR_DIRECTORY" ] && umount "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
+  [ -n "${CHROOT_OUTPUT}" ] && [ -d "${CHROOT_OUTPUT}" ] || return 0
+  [ -z "${_UMOUNT_ALL_ATTEMPTED}" ] || return 0
+  _UMOUNT_ALL_ATTEMPTED=1
+  local mountpoint
+  while IFS= read -r mountpoint; do
+    einfo "Unmounting ${mountpoint} ..."
+    umount -R "${mountpoint}" ; eend $?
```
mika, line 162: AFAICS we're now trying to unmount all mount points without taking care of any preference/ordering in mind, no matter whether it's a parent (`/chroot/`) or children (`/chroot/proc/` + `/chroot/sys/`)  directory, but we don't care?

grml-live
```diff
@@ -151,8 +151,16 @@ fi
...
+  done < <(findmnt -k -rn -o TARGET | awk -v p="${CHROOT_OUTPUT%/}" '$0==p || index($0,p"/")==1')
```
mika: Do I understand this right that we hereby list all currently mounted paths, then only match lines that either match exactly `${CHROOT_OUTPUT}` or start with `${CHROOT_OUTPUT}`  and also de-duplicate e.g. `/chroot` and `/chroot/` into the same directory?

grml-live
```diff
@@ -744,7 +752,8 @@ mkdir -p "$CHROOT_OUTPUT" || bailout 5 "Problem with creating $CHROOT_OUTPUT"
 
 if [ -n "${MIRROR_DIRECTORY}" ] ; then
   mkdir -p "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
-  mount --bind "${MIRROR_DIRECTORY}" "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}"
+  # -o rbind needs util-linux >= 2.39
+  mount -o rbind,ro=recursive "${MIRROR_DIRECTORY}" "${CHROOT_OUTPUT}/${MIRROR_DIRECTORY}" || bailout 5 "Failed to bind-mount ${MIRROR_DIRECTORY}"
```
mika, line 756: nit: this is the only change visible in the git commit message summary, maybe reflect the new unmout/tracking behavior also? (Wouldn't have expected this code to show up when reading the PR/commit message summary :))

(Sorry, I hate GitHub for not being able to comment regarding the commit message at the according place but only elsewhere)

mika (commented):
Interesting idea, thx!

grml-live
zeha: the commit message body says ...

> Drive-by also support layered mounts in MIRROR_DIRECTORY
> and recursive umount cleanup of CHROOT_OUTPUT. Frees us from tracking
> ...

so maybe this is good enough?

grml-live
zeha: what we need is all *direct* children of $CHROOT_OUTPUT that are mountpoints.

unfortunately findmnt has a lot of options, but doesn't help us with this.

$CHROOT_OUTPUT itself shouldn't be a mountpoint. you're right though that if $CHROOT_OUTPUT is a mountpoint, this will break.

grml-live
mika: I'm aware, but in debian/changelog this won't show up then? :)

zeha: right, so that needs gbp-dch: full

zeha: i've split the commits so it clearly shows up in the oneline logs. the findmnt invocation still needs work

grml-live
zeha: the "direct children" thing was a massive think-o on my side

grml-live
zeha: i've reworked this to ignore the parent and umount *all* mountpoints inside CHROOT_DIRECTORY in reverse order. this might not always work, but it shouldn't be worse than before

mika (approved):
Lovely, thanks! (FTR: didn't test/verify it, but LGTM!)

Issue #512, opened 2026-04-29
Remove grml-live debs from deb.grml.org


zeha:
We don't consume grml-live as a Debian package anywhere. We also expect grml-live users to customise the configs, and that they do so using a git branch. Thus, lets:

- [ ] find solution for the ISO changelog
- [ ] update grml-live.8.adoc section `How do I deploy grml-live on a plain Debian installation?` to point to git
- [ ] remove the grml-live(-grml) debs from deb.grml.org
- [ ] remove them from the packages.grml.org list
- [ ] keep building the debian package in github ci for verification

zeha:
Turns out we do ship it in GRML_FULL!

zeha:
We will need a solution for the grml-live changelogs

PR #530, opened 2026-05-29
Update grml-live.8

zeha:
Also addresses #512.

### DIFF
--- a/docs/grml-live.8.adoc
+++ b/docs/grml-live.8.adoc
@@ -297,8 +297,7 @@ The following files and directories are relevant for class GRMLBASE by default:
   ${GRML_FAI_CONFIG}/scripts/GRMLBASE/
   ${GRML_FAI_CONFIG}/media-scripts/GRMLBASE/
 
-Take a look at the next section for information about the concept of those
-files/directories.
+See <<files,Files>> for details about these files and directories.
 
 If you want to use your own configuration, extend an existing configuration
 and/or add additional packages to your ISO just invent a new class (or extend an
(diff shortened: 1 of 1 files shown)


docs/grml-live.8.adoc
```diff
@@ -738,33 +718,40 @@ dpkg-dev is a simple and easy approach.
...
-(after making sure all dependencies are installed), you should set
-`GRML_FAI_CONFIG` so that a) it finds its class-based configuration files and
-b) does not use the config files of an possibly installed `grml-live` package.
 Usage example:
 
-    # export GRML_FAI_CONFIG=$(pwd)/config
-    # ./grml-live -s sid -a amd64 -c GRML_FULL
+    $ git clone --depth 1 https://github.com/grml/grml-live
+    $ sudo apt install <packages>  # see debian/control
+    $ cd grml-live
```
mika, line 737: those lines should be changed though, nor?

(wondering whether there's something like `apt-get build-dep grml-live` or `mk-build-deps` for the runtime depends?)

zeha: changed how? I've left resolving the package list to the reader, due to `Depends` in debian/control is being interpreted by debhelper and dpkg to form a binary package `Depends` line. That needs a full package build.

mika: By switching the lines into:

```
    $ git clone --depth 1 https://github.com/grml/grml-live
    $ cd grml-live
    $ sudo apt install <packages>  # see debian/control
```

Otherwise the file `debian/control` won't be available? :)

zeha: fair. done

mika (approved):
Awesome, thanks!

PR #532, opened 2026-06-14
Cleanup minifai's _main a bit

zeha:
This is work from another branch, that can land now.


zeha:
Builds are failing because of the known systemd-tmpfiles issue in forky (#533).

mika (approved):

Issue #533, opened 2026-06-14
systemd-tmpfiles breaks daily builds


zeha:
```
Setting up sudo (1.9.17p2-7) ...
/proc/ is not mounted, but required for successful operation of systemd-tmpfiles. Please mount /proc/. Alternatively, consider using the --root= or --image= switches.
dpkg: error processing package sudo (--configure):
 old sudo package postinst maintainer script subprocess failed with exit status 1
```

zeha:
Currently we cannot mount `/proc`, as the builds run inside an OCI container (unprivileged of course).
We already tried switching docker for podman, but doesn't help a thing - the container still doesn't have CAP_SYS_ADMIN which is needed for mount.

zeha:
Current plan: do away with the containers altogether, let grml-live run as a normal user, have it create its own userns as needed.

mika:
Do we have a tracking bug in either Debian's BTS or systemd upstream that's related to this failure / behavior change? 🤔

zeha:
> Do we have a tracking bug in either Debian's BTS or systemd upstream that's related to this failure / behavior change? 🤔

No. The last time I reported something similar in another systemd binary, it was closed saying that our build host is broken, effectively claiming that CI with unprivileged containers is not a support target.

mika:
> > Do we have a tracking bug in either Debian's BTS or systemd upstream that's related to this failure / behavior change? 🤔
> 
> No. The last time I reported something similar in another systemd binary, it was closed saying that our build host is broken, effectively claiming that CI with unprivileged containers is not a support target.

😢 

I recall that, but couldn't find the related bug number or alike, do you recall the bug number or so where this happened? :)
(Would be nice to be able to point people to this? IMO this is a problem within systemd :-/)

zeha:
https://github.com/systemd/systemd/issues/38632 was back then

PR #534, opened 2026-06-14
minifai: support APT_PROXY for mmdebstrap

zeha:
Inform mmdebstrap about the proxy for APT, when APT_PROXY is set.


zeha:
Builds are failing because of the known systemd-tmpfiles issue in forky (https://github.com/grml/grml-live/issues/533).

mika (approved):

PR #537, opened 2026-06-24
unshare based builder

zeha:
Produces a working ISO, and GHA CI passes.

There might be hidden dragons.

Known broken / TODO:

- [x] apt mark auto in mmdebstrap produces thousands lines of log output
- [x] iso unpacking
- [x] md5sums file
- [x] log file moving
- [x] CI
- [x] `2026-06-25T16:44:09.6767960Z xorriso : FAILURE : Given path does not exist on disk: -boot_image system_area='/tmp/glbuildwbavewd9/grml_chroot/usr/lib/grub/i386-pc/boot_hybrid.img'`
- [x] ~chroot is now always destroyed~ - need to deprecate some modes and need docs

discarded:
- mksquashfs should show numeric uids only, with name-based user ids the output is very confusing - seems impossible with mksquashfs

changes that were included:
- dropped cap_timestamps before mksquashfs, see #536
- minifai.py became multiple python files
- configdump now also has all variables even if they are empty, was necessary to keep minifai.py simple in its current state
- `GRML_FAI_CONFIG` now gets defaulted to source/config if running from a source checkout

for later:
- parsing the configdump in minifai is brittle - need to see what we need
- buildinfo: moved, but needs big improvement - best have each step record some buildinfo artifact, instead of reconstructing all data when writing the json
- move option parsing from grml-live to a new grml_live.py, merge minifai.py

Closes: #533

PR #543, opened 2026-07-04
RELEASE/98-chroot: run ls -la in target, so ownership is displayed correctly

zeha:
Fixes: f1dc42a964e540edc64d16c9479bcca770368fb6

As noticed in #537


zeha:
Builds fail, but we know that

mika (approved):

PR #545, opened 2026-07-04
grub: make boot_hybrid.img available in ISO9660 filesystem

zeha:
Necessary for remastering, but also for unshare based building when the chroot is already cleaned up when xorriso runs.

Fixes: 3e4c53a595c908647193b130675e1cfc4a054914

Extracted out of #537


zeha:
Builds fail, but we know that

mika (approved):

PR #552, opened 2026-07-12
Misc SecureBoot changes

mika:
* SecureBoot: read secureboot/grub.cfg also from main GRUB configuration
* SecureBoot: fix prefix setting
* SecureBoot: update debugging hints in grub.cfg

Disclaimer: I still don't know why the SecureBoot theme usage broke, `secureboot/grub.cfg` in the efi.img just isn't read any longer. :-/

### DIFF
--- a/config/media-files/GRMLBASE/boot/grub/grub.cfg
+++ b/config/media-files/GRMLBASE/boot/grub/grub.cfg
@@ -1,5 +1,12 @@
 ## grub2 configuration
 
+# secure boot, read only if not already sourced elsewhere
+if [ "${grml_secureboot}" = "" ] ; then
+    if [ -f "${prefix}/../../secureboot/grub.cfg" ] ; then
+        source "${prefix}/../../secureboot/grub.cfg"
+    fi
+fi
+
 if [ -f "${prefix}/grub_local_custom_early.cfg" ] ; then
     source "${prefix}/grub_local_custom_early.cfg"
 fi
--- a/config/media-files/GRMLBASE/secureboot/grub.cfg
+++ b/config/media-files/GRMLBASE/secureboot/grub.cfg
@@ -7,7 +7,7 @@ set grml_orig_root=$root
 export grml_orig_root
 
 search.file %BOOT_FILE% root
-set prefix=($root)/${prefix}/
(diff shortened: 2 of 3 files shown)


config/media-files/GRMLBASE/boot/grub/grub.cfg
```diff
@@ -1,5 +1,12 @@
 ## grub2 configuration
 
+# secure boot, read only if not already sourced elsewhere
+if [ "${grml_secureboot}" = "" ] ; then
```
zeha, line 4: not sure what we really need from `secureboot/grub.cfg`. if we only need `grml_secureboot` being set correctly, maybe its smarter to move the wrmsr test into grub.cfg and not source `secureboot/grub.cfg`?

mika: The `secureboot/grub.cfg` was meant as main entrance point as `/boot/grub/grub.cfg` in the `efi.img` (see `config/media-scripts/GRMLBASE/10-efi`), but yeah, maybe we can re-design this?

zeha: Yeah I remember it was like that, but if it doesn't get loaded, lets put the relevant parts into the normal grub.cfg and keep the (debug) messages in secureboot/grub.cfg?

Sourcing secureboot/grub.cfg from the main config just seems to make things harder to understand?

mika: Absolutely, this only needs further work then. :) FTR, the changes I provide here and in grml-terminalserer as https://github.com/grml/grml-terminalserver/pull/27 are what I was working on for Grml-Forensic, extensively tested and also released as such ("internally"). It's working as-is, the only thing that's not working yet is displaying "SecureBoot" on the splash screen when booting via PXE/netboot, but that's mainly a cosmetic issue only (and I need to see whether there actually is a good solution to get this working at all, given the limitations of test -f ... not working via tftp ;(

(Sorry for the ping-pong in "resolve converstation", GitHub doing strange things on a flaky network connection on a train ride)

zeha (approved):

PR #558, opened 2026-07-20
feature: support execution of grml/main.sh on ESP partition

mika:
Inspired by Marc Haber's "Helping grml to boot noninteractively into network and ssh" mail on grml-devel, see
https://lists.mur.at/hyperkitty/list/grml-devel@ml.grml.org/thread/XKVFTSTYFYDN3JS3ILUFICNQK74O456T/

### DIFF
--- a/config/files/GRMLBASE/usr/share/grml-autoconfig/autoconfig.functions
+++ b/config/files/GRMLBASE/usr/share/grml-autoconfig/autoconfig.functions
@@ -1021,6 +1021,50 @@ elif checkbootparam 'debs' || checkbootparam 'config' || checkbootparam 'scripts
 fi
 }
 
+# Execute grml/main.sh from ESP partition {{{
+config_find_grml_esp() {
+  if ! checkbootparam 'grmlesp' ; then
+    return 0
+  fi
+
+  local mntdir esp_devices
+  mntdir="$(mktemp -d)"
+  # sadly we can't rely on LABEL/PARTLABEL being set to EFI/ESP,
+  # so check for the GUID of an EFI system partition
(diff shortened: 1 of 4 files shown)


config/files/GRMLBASE/usr/share/grml-autoconfig/autoconfig.functions
```diff
@@ -1021,6 +1021,50 @@ elif checkbootparam 'debs' || checkbootparam 'config' || checkbootparam 'scripts
...
+      local grml_script="${mntdir}/${grml_script}"
+
+      if ! [ -r "${grml_script}" ] ; then
+        ewarn "Could not find ${grml_script}"
+        continue
+      fi
+
+      einfo "Executing ${grml_script}"
+      # shellcheck disable=SC1090
+      if . "${grml_script}" ; then
```
zeha: should this really use `.` or run the script in a new shell?

mika: yeah good question, run under `bash`? `/bin/sh`? any suggestion what this should look like?

zeha: or require it to have a shebang and be executable?

mika: good idea! done

zeha:
maybe you can rebase this on current master, to get the CI to run

zeha:
Some thoughts on this:
1) i'd like to avoid having the X-th autoconfig-like thing
2) maybe the boot option can be shared with autoconfig and thus be `noautoconfig` (and default on)
3) cheatcodes.txt needs an update at least, and ideally `docs/grml-autoconfig.current.adoc` (even though i don't see where it currently is on the website) ; (grml.org/config needs cleanup)
4) maybe expect a directory called `GRMLCFG` on the ESP (next to `EFI`, not below it)
5) allow the program to be a non-shell-script program (think go)
6) can we have a clearly defined ENV for the program that runs?
7) can we delete some of the old autoconfig things?

Issue #559, opened 2026-07-20
Rework grml-live behaviour


zeha:
* [x] `-a` / ARCH: i386 is unsupported
* [x] `-a` / ARCH: switching between amd64 and arm64 does not make sense as normally you cannot run binaries from the other arch
* [ ] `-A` - clean artifacts - build-driver refuses anyway; nb: build-driver passes `-A`, needs check if this can be deleted as is
* [x] `-n`: mika says its unused
* [x] `-N`: mika says its unused
* [x] `-Q`: probably can go
* [x] `-U`: was workaround for jenkins, remove
* [x] `-z`: old workaround for squashfs without lzma support, remove

should refactor:
* [ ] `-i`: iso_name is used to derive artifact basename (by stripping `.iso`) - build-driver also(!) constructs all of the artifact names
* [ ] `-b`: description should change to: use existing chroot - currently tells minifai to not install software - used for "repack"! ideally this would run updatebase+instsoft!
* [ ] `-B`: expert option: use existing chroot - run media-scripts, create squashfs, create iso, (no buildinfo) (no scripts)
* [ ] `-q`: expert option: use existing chroot - run media-scripts, SKIP squashfs, create iso, (no buildinfo) (no scripts)

should keep working:
* [ ] `-I`: cleanup help text
* [ ] `-u`: maybe mark as expert option, if it currently works, otherwise remove?

want new option:
* [ ] `--customize-hook` (ala mmdebstrap, for example usecase: evaluate size difference of a new package `--customize-hook=apt install -y foo`)

delete:
* [x] `PRESERVE_LOGFILE`

rework:
* [x] stop making `LOGFILE` settable from outside
* [ ] later rework `LOGFILE` so by default output is only written to logfile; with an option and under CI=true do not redirect output to logfile; ideally have sbuild-style historized logfiles
* [x] stop making `$DPKG_BINARY`, `$MKSQUASHFS_BINARY`, ... settable from outside
* [x] remove NO_ADDONS, mostly pointless
* [x] remove NO_BOOTID
* [x] stop making `SQUASHFS_OPTIONS` settable from outside/config
* [x] stop making `SQUASHFS_EXCLUDES_FILE` settable from outside/config
* [x] move ld aux cache "deletion" into squashfs excludes

other notes:
[...]

PR #563, opened 2026-07-21
Remove NO_ADDONS, NO_BOOTID, SQUASHFS_EXCLUDES_FILE settings

zeha:
Reference: https://github.com/grml/grml-live/issues/559


mika (dismissed):
Maybe we should refer to https://github.com/grml/grml-live/issues/559#issuecomment-5027843102 also in our commits?

zeha:
> Maybe we should refer to [#559 (comment)](https://github.com/grml/grml-live/issues/559#issuecomment-5027843102) also in our commits?

Unsure when there is no reasoning in the issue, but I've added it now.

mika (approved):

PR #565, opened 2026-07-21
GRUB setup: merge scripts/GRMLBASE/45-grub-images into media-scripts 31-grub

zeha:
As media-scripts 10-efi also relied on scripts 45-grub-images, rename that to 39-efi, so it runs after 31-grub.

It would seem we could also merge 10-efi into 31-grub, but this can be done later.


zeha:
This needs verification

mika (commented):
(Untested/unverified, but change itself LGTM, thx!)

mika:
Verified both BIOS + EFI boot with `grml-ghaci-daily20260721build0trixie-amd64.iso`, booting fine!

mika (approved):

PR #567, opened 2026-07-21
Drop support for i386 architecture

zeha:
Removes package lists for i386, removes ifclass/hasclass checks for i386. Removes "i386" to option `-a`.

UEFI 32bit boot on an AMD64 ISO should continue to work.

### DIFF
--- a/build-driver/build.py
+++ b/build-driver/build.py
@@ -395,7 +395,7 @@ def _main(program_name: str, argv: list[str]) -> int:
         usage(program_name)
         return 2
 
-    if arch not in ("amd64", "i386", "arm64"):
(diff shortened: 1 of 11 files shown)


mika (commented):
Maybe also drop the `I386` thingy from `How do I create a base tar.gz (I386.tar.gz or AMD64.tar.gz or ARM64.tar.gz)` in `docs/grml-live.8.adoc`?

zeha:
> Maybe also drop the `I386` thingy from `How do I create a base tar.gz (I386.tar.gz or AMD64.tar.gz or ARM64.tar.gz)` in `docs/grml-live.8.adoc`?

Done!

mika (approved):

PR #571, opened 2026-07-21
grml-live: drop options -n, -N, -Q

zeha:
Removes option -n (SKIP_MKISOFS), -N (FAI_ACTION=bootstrap and skip rest), -Q (SKIP_NETBOOT).

Instead we expect users of these to tolerate the ISO and netboot build times. Specifically for -N, we expect users to run the whole build (incl. the ISO), and only then start customizing the chroot.


zeha:
Includes #570 until it gets merged.

mika (approved):

PR #572, opened 2026-07-21
grml-live: remove -U option

zeha:
Removes option -U (CHOWN_USER), which we have not been using since switching away from Jenkins CI.

https://github.com/grml/grml-live/issues/559


zeha:
`-A` cannot be removed until we switch the build stack and have it always remove the chroot etc.

mika (approved):

PR #576, opened 2026-07-22
grml-live: require -b/-B/-u with existing chroot, update help texts

zeha:
Stop guessing what the user wants:
* with an existing chroot, require one of the -b/-B/-u options.
* without an existing chroot, none of them can be given.
* require that only one of them is ever given at the same time.


mika (dismissed):
Much better, thanks!

PR #578, opened 2026-07-22
grml-live: abort if GRML_FAI_CONFIG is empty/absent

zeha:
Guard against -D pointing to a non-existing directory, as commonly seen when running directly from git.


mika (approved):
🚀

PR #579, opened 2026-07-22
Use squashfs-excludes to ignore ldconfig cache

zeha:
This should allow us reordering mksquashfs and media-scripts running order in the future, if necessary. Also its just nicer.

Fix exclusion of toplevel `/grml-live` directory while we are here.


mika (approved):
Nice catch, thanks! 🙇

PR #580, opened 2026-07-23
grml-live: drop MIRROR_DIRECTORY support

zeha:
The unshare based builder will (at least initially) not be able to support this. Drop the support for now.


mika (approved):

mika:
FTR, forced the merge as the CI failure is related to the perl transition in Debian unstable.

zeha:
> FTR, forced the merge as the CI failure is related to the perl transition in Debian unstable.

thank you!

PR #582, opened 2026-07-23
build-driver: leave grml_cd, grml_chroot alone, stop using -A

zeha:
For our CI we care about the artifacts, not these intermediate, very big directories.

This also allows us to stop using -A in the build-driver.


zeha:
Lets see if the github actions agree.

mika (approved):
🚀

PR #589, opened 2026-07-25
grml-live: remove a noisy debug message

zeha:
The debug message was useful during unshared_helper development, but now its a bit noisy.


mika (commented):
Maybe we could provide those details when running in a debug mode or alike? 🤔

zeha:
this got fixed / less noisy in #635

PR #594, opened 2026-08-02
minifai: fix policy-rc.d cleanup

zeha:
Unlinking was done directly in the minifai process, instead of using the
unshared helper. This leads to a permission error, and the build logs
indeed show:
```
  W: Failed cleaning up /tmp/glbuildjgsbsp6x/grml_chroot/usr/sbin/policy-rc.d
```


zeha:
New build logs look correct.

mika (approved):
Nice catch, thanks!

PR #595, opened 2026-08-02
unshared_helper: fix message framing on socket

zeha:
Previously this could in theory lose messages / have short reads/writes.

Frame the protocol in both directions (not just towards the server), and
retry reading and sending properly.

sendall() fix also applies to the minifai socket thread, so fix it there too.


mika (approved):
uff wow, thx :)

PR #599, opened 2026-08-04
grml-live: replace main script with python

zeha:
**Drops support for configuration file** (grml-live.conf, grml-live.local).
Everything is a command line option now. `SOURCE_DATE_EPCOCH` and `APT_PROXY` stay envvars, by standardisation.

Drops all warnings/rejections for previously deprecated environment/config variables and classes.

Automatic classes can no longer be given in `-c`. Also rejects the old `I386` class.
Doc'd some internal classes (`DEBIAN_*`, `SECURE_BOOT`).

Option `-d` no longer sets `SOURCE_DATE_EPOCH` to the given date. The old implementation felt a bit backwards.

There are further cleanups to be done, but they can also land later. Some are already marked as TODO.


usr/share/zsh/vendor-completions/_grml-live
```diff
@@ -59,24 +59,32 @@ _grmllive_apt_snapshot_dates() { #{{{
 #}}}
 
 arguments=( #{{{
-  '-b[build the ISO without updating the chroot via FAI]'
-  '-B[build the ISO without touching the chroot (skips cleanup)]'
+  '(-w)--bootstrap-mirror-url[Debian mirror URL for creating chroot]:mirror URL:_urls'
+  '--boot-options[add these boot options to the defaults]:boot options:'
+  '--distri-info[set DISTRI_INFO, a one-line description of the built system]:distribution info:'
+  '--distri-name[set DISTRI_NAME]:distribution name:'
```
mika, line 65: is this the same as `-g GRML_NAME` for now?

usr/lib/grml-live/grml_live/main.py
```diff
@@ -0,0 +1,569 @@
...
+        metavar="CONFIGDIR",
+        default=default_paths.config_dir,
+        help="use specified configuration directory",
+    )
+    parser.add_argument(
+        "-e",
+        type=arg_absolute_path,
+        dest="extract_iso_name",
+        metavar="EXTRACT_ISO_NAME",
+        help="extract ISO and squashfs contents from iso_name",
```
mika: this now shows up as:

```
-e EXTRACT_ISO_NAME   extract ISO and squashfs contents from iso_name
```

I guess we should either rename `EXTRACT_ISO_NAME` or `iso_name` to match with each other?

usr/lib/grml-live/grml_live/main.py
```diff
@@ -0,0 +1,569 @@
...
+        default="GRML_FULL",
+        help="classes to be used for building the ISO",
+    )
+
+    parser.add_argument(
+        "--distri-info",
+        type=str,
+        dest="distri_info",
+        default="Grml - Live Linux for system administrators",
+        help="Set DISTRI_INFO",
```
mika, line 129: nit: we should _either_ use `Set DISTRI_INFO` _or_ `set DISTRI_INFO`, but now we'd get a mixture of uppercase vs lowercase in the help text :)

usr/lib/grml-live/grml_live/main.py
```diff
@@ -0,0 +1,569 @@
...
+        type=arg_key_str,
+        dest="grml_version",
+        metavar="VERSION_NUMBER",
+        default="0.0.1",
+        help="set the grml version number",
+    )
+
+    mirror_options = parser.add_mutually_exclusive_group()
+    mirror_options.add_argument(
+        "--bootstrap-mirror-url",
```
mika, line 224: is there any specific reason why this option shows up as next-to-last entry only, while most the rest[1] of the option list is sorted (and displayed) alphabetically (which I like!)?

[1] only exceptions being `--hostname HOSTNAME` + ` --username USERNAME`, maybe we clarify how we want it ordered? :)

usr/lib/grml-live/grml_live/main.py
```diff
@@ -0,0 +1,569 @@
...
+
+  man grml-live
+  /usr/share/doc/grml-live/grml-live.html
+  http://grml.org/grml-live/
+
+Please send your bug reports and feedback to the grml-team: http://grml.org/bugs/
+""",
+    )
+
+    parser.add_argument(
```
mika, line 112: I'm aware that we dropped the `-a` option in previous commit 9a903deb8dc97a8ef1c9d37751ce7bbe1c2aa3c2 already, but we didn't ship a new grml-live release/version/tag with this change yet.

Should we maybe announce the options we no longer support as deprecated and provide according help/usage text? WDYT?

usr/lib/grml-live/grml_live/main.py
```diff
@@ -0,0 +1,569 @@
...
+    )
+    build_type_options.add_argument(
+        "-u",
+        action="store_const",
+        dest="fai_action",
+        const=build_facts.FaiAction.SOFTUPDATE,
+        help="update and build from an existing chroot",
+    )
+    build_type_options.add_argument(
+        "-q", action="store_const", dest="fai_action", const=build_facts.FaiAction.REBUILD_MEDIA, help="skip mksquashfs"
```
mika: AFAICS the `build mode (expert only)` options, being:

```
  -b                    build from an existing chroot
  -B                    build from an existing chroot without running scripts
  -u                    update and build from an existing chroot
  -q                    skip mksquashfs
```

are all conflicting with each other now, like:

```
grml-live: error: argument -B: not allowed with argument -q
grml-live: error: argument -q: not allowed with argument -b
```

AFAICT this seems to match what we discussed and documented at https://github.com/grml/grml-live/issues/559 - right? Maybe we need to improve the help text for `-q` now though, as it's not skipping only squashfs part? 🤔

mika (commented):
Didn't find any actual regressions so far, great work!
Just some minor nitpicks/comments :)

usr/share/zsh/vendor-completions/_grml-live
zeha: No, we expect f.e. `-g grml-small-arm64` and `--distri-name=grml` (default).

usr/lib/grml-live/grml_live/main.py
zeha: The reason is that this option is in an exclusive-group with `-w`, and the exclusive-group is defined later.

If that's a good reason, I don't know.

usr/share/zsh/vendor-completions/_grml-live
mika: Grrr sorry, I clicked at the wrong file (I *hate* GH for code review!).

I meant this:

```
  --distri-name DISTRI_NAME
                        Set DISTRI_NAME (default: grml)
[...]
  -g GRML_NAME          set the grml flavour name (default: grml)
```

So far we didn't have `--distri-name`, nor?

usr/lib/grml-live/grml_live/main.py
zeha: My plan was to replace the `-bBuq` options too and then announce the breakage somewhere. Unclear where though; I guess d/NEWS as previously is not the right place anymore.

usr/lib/grml-live/grml_live/main.py
zeha: yes this is as expected. Better texts welcome, or the new texts can wait until a replacement for all of the options comes.

usr/lib/grml-live/grml_live/main.py
zeha: fixed

usr/lib/grml-live/grml_live/main.py
zeha: fixed

usr/lib/grml-live/grml_live/main.py
mika: Ahhh ok, uff :)

usr/lib/grml-live/grml_live/main.py
mika: Ah that makes sense. Hm, maybe we should start with an actual `Changelog` file with sections like `BREAKING CHANGES` or so? :)

usr/lib/grml-live/grml_live/main.py
zeha: Re ordering in general: I tried ordering the options in some logical way, but at some point I gave up. Especially `--hostname` and `--username` are IMO also problematic and maybe want to be renamed to `--distri-hostname`/`--distri-username` to match the other `--distri...` options, and also avoid clashing with `HOSTNAME` and `USERNAME`.

zeha: > Re ordering in general: I tried ordering the options in some logical way, but at some point I gave up. Especially `--hostname` and `--username` are IMO also problematic and maybe want to be renamed to `--distri-hostname`/`--distri-username` to match the other `--distri...` options, and also avoid clashing with `HOSTNAME` and `USERNAME`.

(Better names even more welcome!)

usr/share/zsh/vendor-completions/_grml-live
zeha: We always had both `DISTRI_NAME` and `GRML_NAME`. But there was never a corresponding `--distri-name` option to set `DISTRI_NAME`; you had to use `grml-live.local` for that.

zeha: See e265e3093a98d723d40b36dde1aba26146aa301f

usr/lib/grml-live/grml_live/main.py
zeha: (To some degree given that `-a` is rejected it should be evident that it doesn't work anymore :) )

usr/share/zsh/vendor-completions/_grml-live
mika: > We always had both `DISTRI_NAME` and `GRML_NAME`. But there was never a corresponding `--distri-name` option to set `DISTRI_NAME`; you had to use `grml-live.local` for that.

Ahhhhh clear now, makes sense - thanks!

zeha:
Generally if the option names are confusing - and I think some are - we can rename them immediately after this lands.

So IMO up for rethinking are:

* `-bBuq` - replace with a form of mode/action/subcommand
* `-g` could use a better name
* `-i` as it influences more things than just the iso's name

usr/lib/grml-live/grml_live/main.py
mika: Probably should have been more explicit, sorry :)

There's a combination of upper case vs down case at more places:

```
  -h, --help            show this help message and exit
  --boot-options DEFAULT_BOOTOPTIONS
                        Add these boot options
  -c CLASS1,CLASS2      classes to be used for building the ISO (default: GRML_FULL)
  --distri-info DISTRI_INFO
                        set DISTRI_INFO (default: Grml - Live Linux for system administrators)
  --distri-name DISTRI_NAME
                        Set DISTRI_NAME (default: grml)
  --hostname HOSTNAME   Set HOSTNAME (default: grml)
  --username USERNAME   Set USERNAME (default: grml)
  -d DATE               use specified date instead of build time as date of release (default: 2026-08-17)
  -D CONFIGDIR          use specified configuration directory (default: /home/mika/src/grml/grml-live/config)
  -e EXTRACT_ISO_NAME   extract ISO and squashfs contents from EXTRACT_ISO_NAME
  -F                    force execution without prompting
  -g GRML_NAME          set the grml flavour name (default: grml)
  -i ISO_NAME           set the name of the resulting ISO (and other build results)
  -I CHROOT_INSTALL_DIRECTORY
                        directory which provides files that should become part of the chroot/ISO
  -o OUTPUT_DIRECTORY   main output directory of the build process (default: /home/mika/src/grml/grml-live/grml)
  -r RELEASE_NAME       set the grml release name (default: grml-live rocks)
  -R                    skip applying the RELEASE class and cleanup
  -s SUITE              Debian suite/release, like: stable, testing, unstable (default: testing)
  --secure-boot         Enable Secure Boot using Debian method
  -v VERSION_NUMBER     set the grml version number (default: 0.0.1)
  --bootstrap-mirror-url BOOTSTRAP_MIRROR_URL
                        Debian mirror URL for creating chroot (default: http://deb.debian.org/debian)
  -w DATE               wayback machine, build system using Debian archives from specified date

build mode (expert only):
  -b                    build from an existing chroot
  -B                    build from an existing chroot without running scripts
  -u                    update and build from an existing chroot
  -q                    skip mksquashfs
```

mika:
> Generally if the option names are confusing - and I think some are - we can rename them immediately after this lands.
> 
> So IMO up for rethinking are:
> 
>     * `-bBuq` - replace with a form of mode/action/subcommand
> 
>     * `-g` could use a better name
> 
>     * `-i` as it influences more things than just the iso's name

Fully agree and perfectly fine for me!

usr/lib/grml-live/grml_live/main.py
zeha: right, i apparently failed at grepping for uppercase Set. have now standardised on starting with a Capital.

mika (approved):
LGTM, thanks! 🙇

PR #602, opened 2026-08-09
format /boot/grub/help/cheatcodes.txt

crpb:
The string which will be replaced by %SQUASHFS_NAME% will differ in length. Print the information in the next line to not see outputs like this:
```
  ...
  toram / grml2ram      copy the whole CD/media to RAM and run from there
  toram=grml-trixie-small.squashfs      copy only specified squashfs file to RAM and
                                                           run from there
  ...
```


mika (approved):
good idea, thanks!

